PR #2, opened 2016-08-08
Bugfix/pd 226 cannot save data to database

gortonz:
Changed the data type of "options" column of database table widgets from jsonb to text as the active admin does not handle the jsonb data type well.


tobyhede:
👍

PR #10, opened 2016-08-12
PD-264 - Added index page into RoR project

gortonz:


catherine-jones:
@gortonz you need to rebase master into your branch and get my fix for the build.

app/helpers/dashboard_helper.rb
```diff
@@ -1,2 +1,34 @@
...
+            else
+              'aus-width-one-half'
+          end
+    return cls
+  end
+
+  def DashboardHelper.truncateDashboardName(name)
+
+  end
+
```
catherine-jones: @gortonz is the helper below being used for anything? I think we could probably remove it

app/helpers/dashboard_helper.rb
```diff
@@ -1,2 +1,34 @@
 module DashboardHelper
+
+  def DashboardHelper.sizeToStyle(size)
+    cls = case size
+            when 'extra-small' then
+              'aus-width-one-fourth'
```
tobyhede: @catherine-jones Is it possible to actually make the CSS more semantic here? So have classes for the charts that set the width based on more human names than 'aus-width-one-fourth'

tobyhede: cc @elisechant

gortonz:
A pull request before Toby adding more back-end stuffs

app/helpers/dashboard_helper.rb
catherine-jones: @tobyhede yeah we could change them that's not hard but not sure what would be more 'human' than what they are already. I was just following what 18F have done.  Suggestions maybe sm-col, thirds?

tobyhede: For charts, couldn't we just use small, medium etc?

catherine-jones:
@goton LGTM

PR #19, opened 2016-08-15
Don't use the Rails Assets Pipeline

elisechant:
we don't want to use the Rails Assets Pipeline (for now), but we still want to use the path helpers

* remove generated files that shouldn't be included in project
* change path helpers


catherine-jones:
LGTM

PR #28, opened 2016-08-17
Pd 318 backport index

catherine-jones:
update index.html and index.js with fixes implemented in productotype.

### DIFF
--- a/app/views/dashboards/index.html.erb
+++ b/app/views/dashboards/index.html.erb
@@ -2,21 +2,18 @@
   <section>
     <div class="dashboards-intro">
       <h1 class="sr-only">Dashboards</h1>
-      <p>The Performance Dashboard makes data open and accessible by measuring the performance of Australian government
-        services
-        against the Digital Service Standard.</p>
-      <p>This promotes government transparency and helps drive the ongoing improvement of government services – for all
-        Australians.
+      <p>The Performance Dashboard makes data open and accessible by measuring the performance of Australian government services against the Digital Service Standard.</p>
+      <p>This promotes government transparency and helps drive the ongoing improvement of government services – for all Australians.
       </p>
     </div>
   </section>
 
   <ul class="list-unstyled clearfix">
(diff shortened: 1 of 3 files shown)


lib/assets/src/scripts/index.js
```diff
@@ -80,9 +80,12 @@ d3.selectAll('.bar .widget__inner').each(function() {
   let dataRaw = widget.datasets;
```
elisechant: @catherine-jones we need to roll out changes to index.js

app/views/dashboards/index.html.erb
```diff
@@ -2,26 +2,22 @@
...
   <ul class="list-unstyled clearfix">
     <% dashboards.each do |dashboard| %>
-      <li class="item--australian-citizenship-appointment-booking-service-dashboard card row">
+      <li class="card row"> <!--TODO add helper to generate class based on dashboard service name-->
         <div class="card__left">
           <div>
-            <h2 class="card-title"><%= dashboard.name %></h2>
+            <h2 class="card-title"><%= dashboard.name %></h2> <!--TODO add helper to ensure dashboard is on the end of the name-->
             <div class="updated-at">
               <p>Updated at <%= dashboard.published_at.strftime("%d %b %Y %H:%M:%S") %></p>
```
elisechant: are we able to store this transformation in a helper?

tobyhede: Yes. We actually have this whole decorator pattern lib that we can use to. `dashboard_decorator.rb`

elisechant: real nice 👍. let's break this off then from this specific task. LGTM

elisechant:
@catherine-jones we should also port the widget partials from that Index route as well? if it's separate task, maybe we need a new ticket for that task?

catherine-jones:
@elisechant yeah i think break it out into another ticket would be good

PR #29, opened 2016-08-17
Add raw data to the overview_widget view as a table

gortonz:
Added the raw data in overview_widget page as a table


elisechant:
@gortonz  is this a duplicate now?

gortonz:
Convert the raw data to a table

PR #30, opened 2016-08-17
PD-321 View helpers

elisechant:


app/controllers/about_controller.rb
```diff
@@ -1,5 +1,12 @@
 class AboutController < ApplicationController
+
+  include AboutHelper
```
elisechant: not required

app/helpers/dashboard_helper.rb
```diff
@@ -1,30 +1,20 @@
...
+               'aus-width-one-third'
+             when 'medium' then
+               'aus-width-one-half'
+             when 'large' then
+               'aus-width-two-thirds'
+             when 'full' then
+               'aus-width-one-whole'
+             else
+               'aus-width-one-half'
+           end
```
catherine-jones: hey @elisechant just wondering why these were changes again? @tobyhede wanted to make them more human

elisechant: We're trying to port the app 1:1. They correlate to Sass class names. If you want to reinvent this wheel it should have it's own task so it's not confusing.

catherine-jones: @elisechant ok but you will have to update the class names one way or another because I don't believe that that set of class names exist anymore.

catherine-jones:
line 16 of index.html.erb - when you rebase with master you will see that I have removed the static class and that we also need a helper to generate the service name as a class

app/helpers/dashboard_helper.rb
elisechant: ok

PR #32, opened 2016-08-18
PD-260 Create show page

tobyhede:
Added the various methods and helpers to ensure the correct wigets can be display in the correct places.


elisechant:
Liking it

PR #39, opened 2016-08-19
Add raw data to the overview_widget view as a table

gortonz:


gortonz:
A minor fix for branch merge

PR #40, opened 2016-08-21
restore dashboard url path to dashboard index widget

elisechant:


elisechant:
@tobyhede wondering if there is actually Rails route helpers that could be used instead?

app/views/dashboards/index.html.erb
```diff
@@ -17,7 +17,7 @@
               <p>Updated at <%= dashboard.published_at.strftime("%d %b %Y %H:%M:%S") %></p>
             </div>
           </div>
-          <a href="<% dashboard_path(dashboard) %>" class="card__link__visible">View <% dashboard.name %></a>
```
elisechant: ahh.. i'm assuming this was the route helper

app/views/dashboards/index.html.erb
```diff
@@ -17,7 +17,7 @@
...
+          <a href="<%= dashboard.dashboard_path %>" class="card__link__visible">View <% dashboard.name %></a>
```
tobyhede: there's a helper. I think i fixed this in my show branch as well

tobyhede:
So I think I fixed this already

elisechant:
ok, closing because no longer relevant

PR #42, opened 2016-08-21
improve get_body_class by changing param from str to array

elisechant:


app/helpers/application_helper.rb
```diff
@@ -2,13 +2,14 @@ module ApplicationHelper
...
   # Params:
-  # - class_names: String of class names to append to class names string
-  def body_class(class_names)
-    class_name_str = self.controller_name + '_' + self.action_name
-    if !class_names.empty?
-      class_name_str << ' ' + class_names
+  # - class_names: Array of additional class names to append to body class
+  # names
+  def body_class(names)
+    class_names = [self.controller_name + '_' + self.action_name]
```
tobyhede: You prolly dont need `self.` here.

app/helpers/application_helper.rb
```diff
@@ -2,13 +2,14 @@ module ApplicationHelper
...
+    if !names.empty?
+      class_names.push(names);
```
tobyhede: Is names a string or array?
If its a string, this is good. 
If array, this will push names onto the end of the array, giving a nested array.

app/helpers/application_helper.rb
```diff
@@ -2,13 +2,14 @@ module ApplicationHelper
...
     end
-    return class_name_str
+    return class_names.join(' ');
```
tobyhede: returns are implicit in ruby, you can drop `return`

app/helpers/application_helper.rb
tobyhede: Something like this should work. 
```
def body_class(names = [])
  class_names = ["#{controller_name}_#{action_name}"] << names
  class_names.flatten.join(' ')
end
```

tobyhede: If names is an array, flatten will ensure that it is a single-dimension array.
If names is empty, flatten just disappears the empty array from the end. 

Ruby has string interpolation if you use double quotes and #{var}.

elisechant:
thanks @tobyhede. Really great Code Review 👍 I'll make your edits today

tobyhede:
Bump

elisechant:
oh yeah !

PR #44, opened 2016-08-22
It converts the markdown data to HTML

gortonz:


app/decorators/dashboard_decorator.rb
```diff
@@ -1,4 +1,56 @@
 class DashboardDecorator < Draper::Decorator
   delegate_all
 
+  @markdown
+
+  def initialize
```
tobyhede: This will override the default initialize method of the decorator, and will have possible side effects.

app/decorators/dashboard_decorator.rb
```diff
@@ -1,4 +1,56 @@
...
+    render_options = {
```
tobyhede: We dont need any of these options, i think

Gemfile
```diff
@@ -27,7 +27,7 @@ gem 'activeadmin', github: 'activeadmin'
 gem 'inherited_resources', github: 'activeadmin/inherited_resources'
 gem 'ransack',             github: 'activerecord-hackery/ransack'
 gem 'draper',              '> 3.x'
-
+gem 'redcarpet', '~>3.3'
```
elisechant: can we do some tab spacing like above please?

app/decorators/dashboard_decorator.rb
```diff
@@ -1,4 +1,56 @@
...
+      strikethrough: true,
+      # will parse superscript after ^, you can wrap superscript in ()
+      superscript: true
+      # will require a space after # in defining headers
+      # space_after_headers: true
+    }
+    @markdown = Redcarpet::Markdown.new(renderer, extensions)
+  end
+
+  def to_html(data)
```
tobyhede: The method can safely assume to convert the dashboard notes field. The dashboard decorator wraps the dashboard model, so this method could just be `@markdown.render(notes).html_safe`.
Return is implicit in ruby, so dont need.

app/decorators/dashboard_decorator.rb
```diff
@@ -1,4 +1,56 @@
 class DashboardDecorator < Draper::Decorator
   delegate_all
 
+  @markdown
```
tobyhede: Don't need to define class attributes, they will be created when first used in your method.

tobyhede:
Rather than initalize, perhaps create a `markdown` method that does the initalize of the redcarpet stuff?

eg 

```
def markdown
  renderer = Redcarpet::Render::HTML.new(render_options)
  @markdown = ... etc etc 
end
```
The above method implicitly returns the markdown object, so when you call `markdown.render(notes).html_safe` in the html method, it is accessing the renderer and markdown objects.

app/decorators/dashboard_decorator.rb
gortonz: Sorry, Toby. I will fix it

spec/decorators/dashboard_decorator_spec.rb
```diff
@@ -14,4 +14,52 @@
...
+2. Second item
+
+And it's easy to create links, like to
+      the [Markdown](http://daringfireball.net/projects/markdown/)
+      page."
+
+      decorator = DashboardDecorator.new()
+      html = decorator.to_html(testData)
+
+      puts "<========= Test Data ======>"
```
tobyhede: This isn't testing anything at the moment, just generating output. Might be easier to take a very simple document and test for expected output. We ca probably assume that the gem itself works, we just make sure we call it correctly? Tomorrow lets pair on writing some assertions

app/decorators/dashboard_decorator.rb
tobyhede: All good. You're learning Ruby and Rails and everything all at once.

app/decorators/dashboard_decorator.rb
gortonz: The options will added the </br> tags to some lines with CR/LF and filter the user-defined tags.

app/decorators/dashboard_decorator.rb
gortonz: Thank you very much, Toby, for making those points

I have changed the code with RoR fashion, not in Java one. Sorry for this. It may take a while for me to get used to RoR but I will certainly do my codes in RoR way.

spec/decorators/dashboard_decorator_spec.rb
gortonz: The test codes are printing out the results for the time-being. I will do some assertions tomorrow or we do them together if you have the time.

app/decorators/dashboard_decorator.rb
tobyhede: I do the opposite thing going back to JavaScript all the time. Wonder why my functions dont work and its because I *haven't* included a return in JS.

gortonz: The programmer life is nowadays a bit hard because we need to deal with multiple programming languages...

gortonz:
Do a pull request and try merging the branch

PR #46, opened 2016-08-23
Filter the dashboard name from the title

gortonz:


app/decorators/dashboard_decorator.rb
```diff
@@ -25,6 +25,16 @@ def notes_to_html
     markdown.render(notes).html_safe
   end
 
+  def filter_dashboard_name title
```
tobyhede: The name field is in the dashboard model and shouldn't be provided as a parameter. the parameter can be handled in the same way as the notes above.

app/decorators/dashboard_decorator.rb
```diff
@@ -25,6 +25,16 @@ def notes_to_html
...
+    idx = title.index "Dashboard"
+    if idx then return title.chomp("Dashboard").strip end
```
tobyhede: The method you want here is probably gsub http://apidock.com/ruby/String/gsub
`name.gsub(/dashboard/i, '!')`

app/decorators/dashboard_decorator.rb
gortonz: will fix it

spec/decorators/dashboard_decorator_spec.rb
```diff
@@ -33,15 +33,38 @@
...
 
     it { is_expected.to eq "<h1>Heading</h1>\n" }
 
   end
 
+  describe "filter dashboard name test" do
+    title = "Australian Citizenship Appointment Booking Service Dashboard"
+    subject { decorator.filter_dashboard_name title }
+
+    it { is_expected.to eq "Australian Citizenship Appointment Booking Service" }
```
tobyhede: 💯

spec/decorators/dashboard_decorator_spec.rb
```diff
@@ -33,15 +33,38 @@
...
 
+  describe 'convert notes to html' do
     subject { decorator.notes_to_html }
 
     it { is_expected.to eq "<h1>Heading</h1>\n" }
 
   end
 
+  describe "filter dashboard name test" do
+    title = "Australian Citizenship Appointment Booking Service Dashboard"
```
tobyhede: You could create a new dashboard here with a name attribute

`let(:dashboard) { FactoryGirl.create(:dashboard, :name => 'Australian Citizenship Appointment Booking Service Dashboard') }`

If you define a new `let` inside a describe block, it will replace the dashboard model defined elsewhere in the spec file. This lets you create instances for specific tests.

PR #50, opened 2016-08-24
To deprecate for Capybara: adds Browser Stack integration to Karma

elisechant:
* requires binaries


tobyhede:
Do we still need this?

tobyhede:
@elisechant What are we doing with this one?

PR #55, opened 2016-08-24
Added a rspec test for dashboard controller

gortonz:
Added a dashboard controller rspec test


tobyhede:
Nice work

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      get :index
+      expect(response).to be_success
+      expect(response).to have_http_status(200)
+      expect(response).to render_template(:index)
+    end
+
+  end
+
+  # Please uncomment the following test when the test data is imported to the test database
+  # describe "GET show page with a valid dashboard id" do
```
tobyhede: You can use the factory to create a dashboard
```
let(:dashboard)    { FactoryGirl.create(:dashboard) }
```

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  #   it "responds successfully with an HTTP 200 status code" do
+  #     get :show, params: {id: 1}
```
tobyhede: And here you can say 
```
params: {id: dashboard.id }
```

spec/controllers/dashboards_controller_spec.rb
gortonz: it does not seem to me to be helpful to use FactoryGirl to create test data in this case. This is because the controller retrieves the data from database, not the mock-up data source...

tobyhede: So the factory actually will create the record in the database, so this should work :)

spec/controllers/dashboards_controller_spec.rb
gortonz: cool! it works, thanks Toby

tobyhede:
@gortonz In general, would prefer if you rebase rather than merge changes. I will add some notes to the readme, but basically:

- fetch the latest changes from master 
- checkout your branch
- run `git rebase master` 
- force push

Need to be careful with force push in case other people are using the branch too, but in general its fine

gortonz:
I am using a github desktop application and it did the merge for me. Next time, I will do the rebase and force push

gortonz:
rebased with master

PR #73, opened 2016-08-26
A few small fixes in index and show pages

gortonz:


app/helpers/dashboard_helper.rb
```diff
@@ -17,4 +17,11 @@ def size_to_style(size)
            end
   end
 
+  def title_end_with_dashboard name
```
tobyhede: This should be pulled into the decorator and its probably worth creating a spec

app/helpers/dashboard_helper.rb
```diff
@@ -17,4 +17,11 @@ def size_to_style(size)
...
+    if !name.downcase.index('dashboard') then
```
tobyhede: Convention in ruby is to drop the 'then'

app/helpers/dashboard_helper.rb
```diff
@@ -17,4 +17,11 @@ def size_to_style(size)
...
+      name += ' Dashboard'
+    end
+    return name
```
tobyhede: There is a built in method on string `name.downcase.include?('dashboard')`

app/helpers/dashboard_helper.rb
gortonz: The page can be hit tens thousands a day and this method gets called many times. I think a page-level helper function (inline embedded code?) saves us a lot of machine clocks... so it gets the page to be rendered quicker than having the decorator.

The rspec is slow. It will slow down CircleCI process if we have too many small unit-tests to be executed with it. This method does nothing than a library call.

app/helpers/dashboard_helper.rb
gortonz: The index method seems to be faster than include? one. Please see this link http://stackoverflow.com/questions/7795160/algorithm-used-in-ruby-for-stringinclude

app/helpers/dashboard_helper.rb
tobyhede: At the moment the page is hit only a handful of times a day, and long term the plan will be to add caching to the application. 

The helper is an object, just like the decorator, there would be very minimal performance difference. Helpers are loaded into the view scope just as a normal object.

The rspec takes 3 seconds to run. All of our code should have tests. This is not negotiable.

PR #81, opened 2016-08-29
Get medicare and marketplace data migrated from node.js app. This is …

gortonz:
…a intermediate submit and there may be a few data and import issues need to be fixed


lib/data/dibp-data.json
```diff
@@ -1,171 +1,177 @@
...
-          { "label": "2016-05", "value":  47 },
-          { "label": "2016-06", "value":  42 }
-      ]
-    },
+        {
+            "id": "device-type-03",
+            "name": "Desktop",
+            "recorded_at": "2016-06-30T01:01:01.111Z",
+            "units": "i",
+            "note": "note for this dataset",
```
elisechant: @mibraheim looks like this probably is lorem ipsum from forever ago?

tobyhede:
That last merge commit merges the branch into itself, I don't think that was the intention?

spec/serializers/widget_serializer_spec.rb
```diff
@@ -41,4 +42,9 @@
     end
   end
 
+  context 'with options' do
```
elisechant: with_options?

app/serializers/widget_serializer.rb
```diff
@@ -2,20 +2,31 @@ class WidgetSerializer < ActiveModel::Serializer
...
 
   def definition
     object.description
   end
 
+  def displayRoundedData?
+    object.options && object.options.key?('displayRoundedData')
+  end
+
   def displayRoundedData
```
elisechant: any reason we have displayRoundedData? and displayRoundedData   ?

tobyhede: One is a check to see if it exists

tobyhede: The other is the field. If you scroll up, its how we make the value conditional ... we add it to the attribute declartion

PR #85, opened 2016-08-31
Bug/pd 423 toggle switch alignment

catherine-jones:
screen shot 2016-08-31 at 3 15 57 pm


lib/assets/src/styles/components/_toggle-switch.scss
```diff
@@ -146,3 +150,21 @@
   width: auto;
   overflow: visible;
 }
+/* ==== Accessibility ===== */
+.aural {
```
elisechant: are you sure we don't have this somewhere else in the code?

lib/assets/src/styles/components/_toggle-switch.scss
```diff
@@ -1,5 +1,6 @@
 #switch-mode {
   width: 58px;
+  margin-right: -69px;
```
elisechant: I think we can probably do better here.

elisechant:
@catherine-jones think this just needs merging with master. Toby changed a bunch of circle stuff yesteday

PR #88, opened 2016-08-31
Update cf-start.sh

davecheney:
Clean up bin/cf-start.sh


bin/cf-start.sh
```diff
@@ -1,4 +1,8 @@
+#!/bin/bash 
+
+set -xeu
```
elisechant: ohhh

PR #91, opened 2016-09-01
Add KPIs and Heros to mygov, turn them off.

tobyhede:
Resolves both PD-441 and PD-440. 

PD-440 requires the fix in PD-441.


elisechant:
looks legit. I need to pull it down and run it locally. I'll do that just after lunch.

gortonz:
We added the KPIs data into mygov-data.json file. It does not seem to have those KPI data in the original javascript data file. Should we always add them next time if they are not in javascript data file?

tobyhede:
@gortonz The way that I am handling the KPIs means that I assume they are the first row. If we have them and don't publish them , it resolves the issue. 

We don't have any new dashboards being released in the next week, so should not need to update.

tobyhede:
YOLO

PR #104, opened 2016-09-02
Fix/dashboardnamethingthatwasannoyingme

tobyhede:


elisechant:
hahahahah.. funniest branch name ever ! @tobyhede you are hilarious :)

app/controllers/dashboards_controller.rb
```diff
@@ -4,7 +4,7 @@ class DashboardsController < ApplicationController
   helper_method :dashboards, :dashboard
 
   def index
-    @dashboards = Dashboard.published.by_name.all
+    @dashboards = Dashboard.published.by_name.all.decorate
```
elisechant: good call

app/decorators/widget_decorator.rb
```diff
@@ -33,8 +33,8 @@ def summary
     end
   end
 
-  def humanised_last_updated
-    last_updated_at.to_formatted_s(:day_month_year).strip
+  def last_updated_at
+    object.last_updated_at.to_formatted_s(:day_month_year).strip
```
elisechant: what is "object"?

tobyhede: object is the model that the decorator is decorating. 
The convention is to wrap the method name rather than create a new name

PR #112, opened 2016-09-05
Moved size_to_style method from helper to decorate class

gortonz:


app/decorators/widget_decorator.rb
```diff
@@ -63,4 +63,15 @@ def name_slug
     name.downcase.parameterize()
   end
 
+  # create a class level hash to store the size to style mapping data for a faster lookup than case-when statement
+  @@sizeToStyleHash = {'extra-small' => 'aus-width-one-fourth', 'small' => 'aus-width-one-third',
```
tobyhede: You can make this a constant and it will read clearer. Constants in Ruby are just all-caps.

SIZE_TO_STYLE =  ...

app/decorators/widget_decorator.rb
```diff
@@ -63,4 +63,15 @@ def name_slug
...
+                       'medium' => 'aus-width-one-half', 'large' => 'aus-width-two-thirds',
+                       'full' => 'aus-width-one-whole'}
+
+  def size_to_style
+    style = @@sizeToStyleHash[size]
+    return style unless style == nil
+    return 'aus-width-one-half'
```
tobyhede: Remember that Ruby has an `||` for booleans. 

So assuming that we use a constant for the hash

```
def size_to_style
  SIZE_TO_STYLE[size]  || 'aus-width-one-half'
end
```

Hash will return nil if the size cannot be found, nil is false.

app/decorators/widget_decorator.rb
tobyhede: Ruby style would be to use a constant here. 
We pretty much never see @@ class variables in the wild.

spec/decorators/widget_decorator_spec.rb
```diff
@@ -482,5 +482,16 @@
     end
   end
 
+  describe 'size extra-small to style' do
```
tobyhede: One thing I just thought of. In dashboard_decorator I refactored the variables to use an options hash. 

`  let(:opts)      { {} }`

Then in each test you can just set the options to have your expectation:

` let(:opts) { { :display_hero => false, :display_kpis => false } }`

Size could be pulled out here with the same pattern?
Does that make sense?

PR #114, opened 2016-09-05
fix - temp remove unrequired css

elisechant:


catherine-jones:
hurray that was driving me mad 💯

PR #132, opened 2016-09-07
conditionally load high contrast

catherine-jones:


elisechant:
Closing because not needed any more.

PR #135, opened 2016-09-08
Feature/pd 476 download dashboard data in csv

gortonz:


config/environment.rb
```diff
@@ -3,3 +3,5 @@
 
 # Initialize the Rails application.
 Rails.application.initialize!
+
+ENV["RAILS_ENV"] = "development"
```
tobyhede: Please remove. This will effect all environments.

app/controllers/dashboards_controller.rb
```diff
@@ -11,10 +11,19 @@ def index
...
     @description = @dashboard.name
 
     render :show
   end
+
+  def export
+    @dashboard = Dashboard.find(params[:id]).decorate
+
+    respond_to do |format|
+      format.html
```
tobyhede: Not required

spec/decorators/dashboard_decorator_spec.rb
```diff
@@ -19,20 +19,20 @@
   end
 
   context 'hide hero' do
-    let(:opts) { { :display_hero => false } }
+    let(:opts) { {:display_hero => false} }
```
tobyhede: Unsure why all of these values have changed?

app/decorators/dashboard_decorator.rb
```diff
@@ -37,7 +37,77 @@ def last_updated_at
...
+
+    end
+
+    # uncomment following for debugging only
+    # puts cnt.to_s
+    # return cnt
+  end
+
   private
+  CVS_COL_NAME = ['id', 'last_updated_at', 'name','type','size','units','description', 'options', 'data_label', 'data_value']
```
tobyhede: I think the format doesn't need the widgets, and it does need the dataset values:

`[dataset_id, dataset_name, timestamp, value]`

tobyhede: You shouldn't be using the rows here, as this is tied to the layout and makes the code complicated. You can just iterate through the `dashboard.widgets`

spec/decorators/dashboard_decorator_spec.rb
gortonz: This could be something done by a code formatter

app/decorators/dashboard_decorator.rb
gortonz: Good ideas, make it simpler... will change it later

tobyhede: I would probably pull all of this into its own object. It doesn't really belong in the decorator as its not tied to the view.

app/decorators/dashboard_decorator.rb
```diff
@@ -44,7 +44,7 @@ def to_csv
       self.widgets.each do |widget|
         dataset = widget.dataset
```
tobyhede: Widgets can have more than one dataset.

app/decorators/dashboard_decorator.rb
```diff
@@ -37,7 +37,29 @@ def last_updated_at
     object.last_updated_at.to_formatted_s(:month_year)
   end
 
+  def to_csv
+    CSV.generate({}) do |csv|
+      csv.add_row CVS_COL_NAME
+
+      self.widgets.each do |widget|
+        dataset = widget.dataset
+        if !dataset.nil? && !dataset.datapoints.nil?
```
tobyhede: A dataset cannot be nil, a widget either has datasets or it doesn't 
widget.datasets is an active record association and returns [] 

The same applies to dataset.datapoints. If is is empty, iterating is a no-oop
`dataset.datatpoints.collect` will return an empty array.

app/decorators/dashboard_decorator.rb
```diff
@@ -37,7 +37,29 @@ def last_updated_at
...
+          darray = dataset.datapoints.collect { |d| [dataset.name, dataset.units, d.ts.strftime("%d/%m/%Y %H:%M"), d.label(), d.value.to_s] }
+
+          if darray.any?
+            darray.each do |item|
```
tobyhede: Again, each is a noop on an empty array.

app/decorators/dashboard_decorator.rb
```diff
@@ -37,7 +37,29 @@ def last_updated_at
...
+              csv.add_row item
```
tobyhede: There should be a way to push the whole array structure on to the csv without iterating through the array again.

Alternatively, don't use `collect`, use an each and append

```
dataset.datapoints.each { |d| csv << [dataset.name, dataset.units, d.ts.strftime("%d/%m/%Y %H:%M"), d.label(), d.value.to_s] }
```

app/decorators/dashboard_decorator.rb
```diff
@@ -37,7 +37,22 @@ def last_updated_at
     object.last_updated_at.to_formatted_s(:month_year)
   end
 
+  def to_csv
+    CSV.generate({}) do |csv|
+      csv.add_row CVS_COL_NAME
+
+      self.widgets.each do |widget|
+        datasets = widget.datasets
+        datasets.each do |dataset|
```
tobyhede: Don't need the assignment here, could just be `widget.datasets.each`

app/decorators/dashboard_decorator.rb
```diff
@@ -37,7 +37,22 @@ def last_updated_at
...
+          dataset.datapoints.each { |d| csv << [dataset.name, dataset.units, d.ts.strftime("%d/%m/%Y %H:%M"), d.label(), d.value.to_s] }
```
tobyhede: How does the timestamp look without formatting? Do we need to do this? 
And Rails has a built in formatter with some nice helpers, see http://apidock.com/rails/ActiveSupport/CoreExtensions/DateTime/Conversions/to_formatted_s

config/routes.rb
```diff
@@ -8,4 +8,7 @@
   get '/index.html', :to => redirect('/')
 
   get '/copyright', :to => 'about#index'
+
+  get '/export', :to => 'dashboards#export'
```
tobyhede: Method should be on dashboards resource 

`/dashboards/id/export`

Rails has a built in way of doing this:
http://guides.rubyonrails.org/routing.html#adding-more-restful-actions

```
resources :dashboards, :only => [:index, :show] do
  get :export
end
```
This then makes a helper available in the view for generating the `export_dashboard_path`

spec/serializers/dashboard_cvs_serializer_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(parser[0].length).to be == 5 # data should be with 5 columns
+      # parser.each do |row|
+      #   puts row.to_s
+      # end
+    end
+  end
+
+  describe "dashboard without widgets test" do
+    let(:empty_dashboard) { FactoryGirl.create(:dashboard, opts) }
+    subject(:serializer2) { DashboardCSVSerializer.new(empty_dashboard) }
```
tobyhede: Can leave out subject here, but change the definition of dashboard. The variable will be used to populate the subject on line 8

` let(:dashboard) { FactoryGirl.create(:dashboard, opts) }`

spec/serializers/dashboard_cvs_serializer_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it do
+      csv = serializer2.to_csv
+      parser = CSV.parse(csv)
+      expect(parser).not_to eq be_empty
+      expect(parser.length).to be == 1  # it should contain a csv header line
```
tobyhede: Could actually check that the header has the right fields.

spec/serializers/dashboard_cvs_serializer_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  describe 'data format test' do
+      # This test case should be failed if the data format from the DashboardCSVSerializer.csv method is not CSV
+    it do
+      csv = serializer.to_csv
+      parser = CSV.parse(csv)
+      expect(parser).not_to eq be_empty
+      expect(parser.length).to be >= 1  # it should, at least, have a csv header line
+      expect(parser[0].length).to be == 5 # data should be with 5 columns
```
tobyhede: rspec has lots of helpers for making assertions about strings and collections

`expect(csv).to include(dataset.name)`

spec/decorators/dashboard_decorator_spec.rb
tobyhede: Are you running one?

app/controllers/dashboards_controller.rb
tobyhede: ^^ This controller doesn't return html, can be removed.

app/serializers/dashboard_csv_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  CVS_COL_NAME = ['dataset_name', 'units', 'time_stamp', 'label', 'value']
+  def to_csv
+    CSV.generate({}) do |csv|
+      csv.add_row CVS_COL_NAME
+
+      @dashboard.widgets.each do |widget|
+        datasets = widget.datasets
+        datasets.each do |dataset|
```
tobyhede: See my previous feedback

config/routes.rb
gortonz: This does not seem to work for me. The way I have done is from rails document but may not be the popular one. This sort of work takes us time but nothing really gains from the changes...

elisechant: +1 `/dashboards/id/export`

spec/decorators/dashboard_decorator_spec.rb
gortonz: I am not running one, I may press an key for code formatter accidentally in rubymine.

spec/serializers/dashboard_cvs_serializer_spec.rb
gortonz: Cool! it works magically. how dose the rspec program flow, from line #32 then jump to line #8?

spec/serializers/dashboard_cvs_serializer_spec.rb
gortonz: definitely can be done

PR #137, opened 2016-09-08
soft release introduce new data structure for front end

elisechant:


elisechant:
requires PD-500 to be merged first.

app/helpers/application_helper.rb
```diff
@@ -1,10 +1,17 @@
...
+  end
+
   # Sets the body class on a per-page basis
   # Params:
   # - class_names: String of class names to append to class names string
   def body_class(class_names)
-    class_name_str = self.controller_name + '_' + self.action_name
+    class_name_str = body_route
     if !class_names.empty?
       class_name_str << ' ' + class_names
```
tobyhede: Could probably drop the assignement and just ` body_route <<`

app/views/layouts/application.html.erb
```diff
@@ -24,7 +22,7 @@
   <%= render 'layouts/analytics' %>
 </head>
 
-<body class="<%= body_class(yield(:body_class_names)) %>">
+<body class="<%= body_class(yield(:body_class_names)) %>" data-route="<%= body_route %>">
```
tobyhede: This smells funny. 
You should be able to reuse content_for in the helper

```
def body_class
  class_names = content_for(:body_class_names)
```

tobyhede:
What does this actually do?

app/helpers/application_helper.rb
elisechant: yep, that's in the other ticket. Getting to it

app/views/layouts/application.html.erb
elisechant: i only want the route here, not all the classes like high_contrast_mode etc

elisechant:
@tobyhede porting the gulp linear javascript in to webpack land so we can work with it (slightly better structured but no dramatic change to original codes)

PR #141, opened 2016-09-13
Minor doc/readme update

joahua:
One minor and one trivial thing I hit when trying to get the dashboard up and running the other night.

Firstly, we should call out the Node dependency prior to referencing npm etc. (TBH I mostly noticed as my laptop had a weird old version that was failing for other reasons). I've just referenced the version used in `package.json` but there might be a better way to do this / more explicitly supported version we should use?

Secondly, I tried to read the Medium link and it 404'd as there was a stray character at the end of the URL.

Apologies for the default `patch-1` branch name also, I used Github's web interface as don't have certs, git, etc., setup on my DTO laptop yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Currently WIP.
 
  - ruby-2.3.1
  - Postgres 9.4
+ - Node 6.3.0
 
 
 ## Getting Started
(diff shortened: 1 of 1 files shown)


elisechant:
@joahua caveat the app probably wont work on PC

joahua:
Indeed, tis great Rails-y folly to try. Though vagrant and docker and stuff
I suppose are options.

elisechant:
@joahua Vagrant would be an excellent idea. At the moment all our devs are on Mac so this hasn't been a huge issue. Basically vagrant should be mandated across DTO. It's really up to @tobyhede what to do there.

elisechant:
Best solution, buy a Mac.

PR #142, opened 2016-09-13
Feature/pd 526 add devise gem and setup basic user login

gortonz:
Added devise auth-library for dashboard authentication functions


tobyhede:
Lots of code is just commented out, is it worth removing the elements from the UI, but keeping the server side code intact?

tobyhede:
Actually. You could leave the classes and hide the routes in production.

gortonz:
Are we going to implement additional authentication logic on top of Devise? I will remove the Controllers from the source if we will not do it.

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,18 @@
 class ApplicationController < ActionController::Base
 
-  http_basic_authenticate_with :name => ENV['HTTP_BASIC_USERNAME'], :password => ENV['HTTP_BASIC_PASSWORD'] if (Rails.env.cloudfoundry? || Rails.env.staging?)
+  #http_basic_authenticate_with :name => ENV['HTTP_BASIC_USERNAME'], :password => ENV['HTTP_BASIC_PASSWORD'] if (Rails.env.cloudfoundry? || Rails.env.staging?)
```
tobyhede: Please do not remove the basic auth

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    super
+  end
+
+  # POST /resource/sign_in
+  def create
+    super
+  end
+
+# DELETE /resource/sign_out
+  # def destroy
```
tobyhede: We probably want the sign_out path?

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,18 @@
...
 
+  # The following several lines should be moved into the controllers need to be authenticated
   protect_from_forgery with: :exception
+  before_action :configure_permitted_parameters, if: :devise_controller?
+  before_action :authenticate_user!, except: [:index, :show, :export]
```
tobyhede: This will effect every controller that extends Application controller. 

The auth methods should be moved to the portal controller. This is the only controller that needs authentication.

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,18 @@
...
+
+  protected
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:sign_up, keys: [:user_name])
```
tobyhede: We can remove all the signup related stuff.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
```
tobyhede: Class not needed.

app/controllers/application_controller.rb
gortonz: Agree with you about this (please see my comments in the code). I will move it the portal controller when it is created

tobyhede: Portal controller is now in master

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Users::PasswordsController < Devise::PasswordsController
```
tobyhede: Lets handle this at the route level and not commit commented out code.

app/controllers/users/unlocks_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Users::UnlocksController < Devise::UnlocksController
```
tobyhede: Same as Password above.

app/controllers/portal_controller.rb
```diff
@@ -1,5 +1,15 @@
 class PortalController < ApplicationController
 
+  before_action :configure_permitted_parameters, if: :devise_controller?
+  before_action :authenticate_user!, except: [:index]
```
tobyhede: We want the index to be authenticated. 
I think a spec would be good for the portal controller. 

Only logged in users should be able to access.

db/migrate/20160912233231_devise_create_users.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      ## Trackable
+      t.integer  :sign_in_count, default: 0, null: false
+      t.datetime :current_sign_in_at
+      t.datetime :last_sign_in_at
+      t.inet     :current_sign_in_ip
+      t.inet     :last_sign_in_ip
+
+      ## Confirmable
+      # t.string   :confirmation_token
```
tobyhede: We can add these fields now if we are going to use them once the email is set up.

app/controllers/portal_controller.rb
```diff
@@ -1,5 +1,15 @@
...
+
+  protected
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:account_update, keys: [:user_name])
```
tobyhede: Do we need this? What does it actually do?

app/controllers/users/passwords_controller.rb
gortonz: Not sure what your meaning of "handle this at route level". I thought that this task is just intermediate one and we will add more auth features later on. I will remove those unwanted codes if we do not need them

tobyhede: We should be able to turn off password reset by disabling the route access. 
Rather than commenting out code. 
I will investigate

app/controllers/portal_controller.rb
gortonz: We may not really need them if we do not allow user to update their account...

tobyhede:
I added confirmable because I think we will need it when we manually create users.
Email should be available early next week. I think we roll with this for now, and add the email config once we have the settings.

PR #146, opened 2016-09-14
PD-502 update dashboard form

elisechant:


client/src/portal/scripts/api/dashboard.js
```diff
@@ -0,0 +1,29 @@
...
+  if (USE_FIXTURES) {
+    return new Promise((resolve, reject) => {
+      setTimeout(() => {
+        resolve({data, status:'success'});  // success interface
+        // reject();                           // failure interface
+      }, 800);
+    });
+  } else {
+    // todo - token etc
+    return fetch(`/api/dashboard/${id}`, {
```
tobyhede: Full path will have a version identifier. Maybe the path is also a constant?

`/api/v1/dashboards/${id}`

PR #157, opened 2016-09-19
Fix/pd 533 layout and design login password pages

gortonz:


app/views/devise/shared/_links.html.erb
```diff
@@ -8,12 +8,9 @@
 <% end -%>
 <% end %>
 
+<% if false %>
```
elisechant: if false ??

app/views/users/sessions/new.html.erb
```diff
@@ -1,26 +1,40 @@
...
-  </div>
-<% end %>
-
-<%= render "devise/shared/links" %>
+<%
+  content_for :title, 'User Login Form'
+  content_for :description, 'Performance Dashboard Login Form'
+%>
+
+<!--<link rel="stylesheet" type="text/css" href="https://gov-au-ui-kit.apps.staging.digital.gov.au/latest/ui-kit.min.css"/>-->
```
elisechant: ?

app/views/devise/shared/_links.html.erb
gortonz: just a comment out in the erb file

PR #158, opened 2016-09-19
PD-538 - common uikit header, footer, layout

elisechant:
and: 
* redirect from update page
* api middleware next loop not dispatch


elisechant:
@catherine-jones i'm starting to put ui-kit stuffs in portal. we should discuss when you're in.

PR #174, opened 2016-09-22
Apparently yield is not nil when empty

tobyhede:


tobyhede:
I SWEAR I TESTED THIS

PR #180, opened 2016-09-25
PD-584 - implements client route transitions in portal

elisechant:
add route transitions;

match route transition with scroll top behaviour

make route transition faster and remove debugging css;

improve timing

abstract animation vals; fix cancel datapoint action; temp fix fixed height header;

tighten up animation speed


client/src/portal/styles/layouts/_app-layout.scss
```diff
@@ -0,0 +1,90 @@
...
+  width: 100%;
+}
+
+body {
+  position: relative;
+}
+
+.l-top-bar {
+  height: #{69 +52}px;
+  @media (min-width: $screen-sm) {  // todo - create a variable to encapsulate this - or use flexbox !!
```
elisechant: or use fixed layout

PR #195, opened 2016-09-29
PD-537 dynamic portal index data

micapam:
This is just a WIP of dynamic JSON for portal/index, based on @elisechant's branch. 

I have followed the suggested JSON structure in the ticket (PD-537), but it doesn't seem to be rendering correctly. Maybe there are some extra attributes that need to be added? 

Screenshot of JSON: 

image

Screenshot of failure of this data to be displayed: 

image

... I'm clearly missing something!


tobyhede:
Ah ... I think I know the problem. Elise had just done a short-hand for the structure, capturing the high level, and ids. I knew what I meant when I wrote the ticket for myself :) 

So we need the actual data attributes too, following what the example is in the code, rather than the ticket. The current json structure in the portal index actually loads. So match that.

app/views/portal/index.html.erb
```diff
@@ -1,4 +1,28 @@
 <script type="text/javascript">
+  window.__CONFIG__ = {
+    OPTIONS_WIDGET_TYPE: [
```
elisechant: can we store these in the database @micapam  ?

elisechant:
@micapam  also, what's the plan for integration? do i merge first?

app/views/portal/index.html.erb
micapam: We could do, or in a YAML file if they aren't going to change much / are global?

micapam:
Yes @elisechant I'll wait for you to merge.

app/controllers/portal_controller.rb
```diff
@@ -1,12 +1,14 @@
...
   before_action :authenticate_user!
   helper_method :session_token
 
   protected
 
   def index
-    render :index
+    respond_to do |format|
+      format.html
+      format.json
```
tobyhede: Do we need the JSON?

app/views/portal/index.json.jbuilder
```diff
@@ -0,0 +1,31 @@
...
+end
+
+json.datasets current_user.datasets do |dataset|
+  json.(dataset, :id, :name, :label, :units, :notes)
+  json.datapoints dataset.datapoints.collect {|datapoint| datapoint.id }
+end
+
+json.datapoints current_user.datapoints do |datapoint|
+  json.(datapoint, :id, :value, :label)
+  json.ts datapoint.ts.strftime('%Y-%m-%d %H:%M:%S.%L')
```
tobyhede: Is it worth adding this to `config/time_format.rb`?

app/controllers/portal_controller.rb
micapam: Hmm maybe not, actually.... since the jbuilder view is only called by the HTML view (currently). I'll try without & make sure it's all good.

app/views/portal/index.json.jbuilder
micapam: I wasn't aware this existed :)

PR #205, opened 2016-10-04
Create an API user for Sandbox envs

tobyhede:
Bootstrap a new user with token access to a nominated dashboard using data provided by ENV variable. 

Various cleanups for the admin area.


.codeclimate.yml
```diff
@@ -12,6 +12,8 @@ engines:
       languages:
       - ruby
       - javascript
+    exclude_fingerprints:
```
tobyhede: Exclude the "duplicate" active admin expire and display token methods.
The effort to remove the duplicate code is greater than the cost of having the dupe imo

lib/tasks/bootstrap.rake
```diff
@@ -1,10 +1,30 @@
...
       user = User.new(:email => ENV['ADMIN_EMAIL'], :password => ENV['ADMIN_PASSWORD'], :password_confirmation => ENV['ADMIN_PASSWORD'])
       user.skip_confirmation!
       user.save!
     end
   end
+
+  task :sandbox_user => :environment  do
+
+    if ENV['SANDBOX_USER']
+      data = JSON.parse(ENV['SANDBOX_USER'])
```
tobyhede: Blat the data in as a JSON blob. Ugly, but creating individual vars is tedious

PR #207, opened 2016-10-04
PD-567 - change 'portal' to 'editor' - client-side

elisechant:


elisechant:
screen shot 2016-10-04 at 3 21 49 pm

PR #209, opened 2016-10-04
PD-598 - include the breadcrumb for all views

elisechant:


elisechant:
screen shot 2016-10-04 at 1 41 35 pm

PR #220, opened 2016-10-05
PD-546 - add "login" layout and asset group; edits form markup;

elisechant:


elisechant:
screen shot 2016-10-05 at 1 39 47 pm

elisechant:
screen shot 2016-10-05 at 2 17 22 pm
screen shot 2016-10-05 at 2 17 00 pm

app/views/users/passwords/new.html.erb
```diff
@@ -0,0 +1,31 @@
+<%
```
elisechant: @micapam @tobyhede added this and `app/views/users/unlocks/new.html.erb`, how can I magically hook them up like ` app/views/users/sessions/new.html.erb` is hooked up?

elisechant: @tobyhede ok, assume we're not intending the separation then.

tobyhede: @elisechant What do you mean? The devise files are for our customisation.

elisechant: app/views/users/sessions/new.html.erb app/controllers/users/sessions_controller.rb

elisechant:
will change http://localhost:3000/login to http://localhost:3000/sign-in based off govau

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Users::PasswordsController < Devise::PasswordsController
+
+  layout 'login'
```
micapam: Does this layout apply to all Devise controllers? If so you can just rename your `login` layout to `devise` as this will automatically be picked up: https://github.com/plataformatec/devise/wiki/How-To:-Create-custom-layouts

app/views/layouts/login.html.erb
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Login - Performance Dashboards</title>
```
micapam: Maybe this wording should also be 'Sign in' as per wording in _links.html.erb

app/controllers/users/passwords_controller.rb
elisechant: ahhh..

app/views/layouts/login.html.erb
elisechant: yes

PR #221, opened 2016-10-05
Allow bulk create of datapoints

tobyhede:
Create endpoint now accepts array as parameter`:datapoints => [{:ts => ts, :value => value}]`

Still handles single datapoint as well


app/controllers/api/v1/datapoints_controller.rb
```diff
@@ -15,15 +15,15 @@ def show
 
   def create
     with_invalid_record_handler do
-      datapoint = dataset.datapoints.create!(data)
+      datapoint = dataset.datapoints.create!(extract_datapoints)
```
tobyhede: @micapam I am not 100% on the approach here for extracting the data from params.
If we have an array of `datapoints`, extract the `[]`, otherwise assume its a flat model.
Active Record will actually take an array of params and turn into a single insert, which is slick. 
Any thoughts?

PR #223, opened 2016-10-06
fix - 'Last updated' date is not displaying on service dashboards mobile

elisechant:


client/src/dashboard/styles/dashboard/_dashboard.scss
```diff
@@ -300,3 +300,12 @@
     }
   }
 }
+
+
+.dashboard__below-the-line {
+  .has-data-true {
+    .widget__last-updated {
+      display: block !important;
```
catherine-jones: is their anyway of doing this without the important?

elisechant: Don't know because i don't understand the css from a distance. Lowest risk bug fix I could come up with.

PR #235, opened 2016-10-10
PD-514 make dataset names more meaningful

tobyhede:


lib/data/dibp-data.json
```diff
@@ -57,7 +57,7 @@
 
         {
             "id": "device-type-01",
-            "name": "Mobile",
+            "name": "Devise Usage - Mobile",
```
elisechant: I feel like an account admin would probably mess this convention up in about 2 minutes

tobyhede: They can do what they want, as long as it makes sense to them.

tobyhede: I feel like we will ultimately need something like tags so that people can find datasets

PR #238, opened 2016-10-11
Feature/devise 2fa

micapam:
UX needs to be thought about - I have got this working though.


tobyhede:
Not sure what the best approach is ... was the consensus that switching gems is the best idea?

micapam:
I've gone back to the Houdini gem, and working with Phil on the UX, so parking this.

PR #241, opened 2016-10-11
PD-636 - Implement state management tests (squash)

elisechant:


elisechant:
@catherine-jones doing some testing here for Editor, yesterday you asked about it, maybe you can take a quick look at what's going on in here and we can do a proper walk through once i've gotten a bit further along ?

PR #256, opened 2016-10-13
Hotfix/pd 671 widget layout firefox

catherine-jones:
screen shot 2016-10-13 at 5 28 00 pm
really horrible grid overrides


client/src/dashboard/styles/_layout.scss
```diff
@@ -44,3 +44,79 @@
...
+      [class*='aus-width-']  {
+        &:nth-child(even) {
+          margin-right: 0;
+        }
+      }
+    }
+  }
+}
+
+// end horrific hack
```
elisechant: wow !

elisechant: why don't we consider using the bootstrap grid here instead? talk tomorrow

catherine-jones: I think we would be better off using something from here http://refills.bourbon.io/

tobyhede: That seems complicated

tobyhede: The original code, not the bourbon idea.

catherine-jones: on tablet it takes the last odd column with a widget in it and make it full width instead of half if we use one of the fluid Bourbon flex grids it should be a big improvement would need to think about fallbacks though. I think reassessing the behaviors we have and might have and then architecting a solution would be a good idea.

client/src/dashboard/styles/_layout.scss
```diff
@@ -44,3 +44,79 @@
     }
   }
 }
+
+// tablet grid hack TODO refactor grid into fluid layout
+
+.medicare-newborn-enrolment {
+  .dashboard__below-the-line {
```
elisechant: The code smell is here. It's great that we're using layout components to abstract layout from presentation, but this code ties the layout back in to a presentation concern. A better solution would be to create new extension classes for layout that handled this media query condition.

client/src/dashboard/styles/_layout.scss
elisechant: hey @catherine-jones , I get what you're trying to achieve, but I agree I think as this _is not a small patch_ (not hotfix) it needs to be rearchitected. We definitely need a grid component that is distinct from presentation components. I have a strong opinion that that should be Bootstrap like we've done in the Editor.  How about we stop work on this for now until we regroup this morning. Focus on small patch bug fixes instead, until we decide what the solution for this bug should be.

tobyhede:
Looks like this may be an obsolete PR< pending refactor and addition of bootstrap grid?

client/src/dashboard/styles/_layout.scss
catherine-jones: one of the reasons that we don't use bootstrap is because of this style of polluting the markup with classes.

elisechant:
Please do not merge.

PR #273, opened 2016-10-14
PD-675 - grid component

elisechant:
* adds bootstrap grid
* sets old 'container' classes to 'container-ish' so no clobber
* adds clearfix to 'container-ish'


tobyhede:
LGTM, although is 'ish' understandable?

elisechant:
Close because cherry picked else where.

PR #279, opened 2016-10-17
Bug fix/pd 631 accessibility 2

catherine-jones:
performance dashboaracc


catherine-jones:
- [ ] check on  iphone 6
- [ ] check on  iphone 7
- [ ] check on  ipad retina 6
- [ ] check on  ipad pro
- [ ] check on firefox
- [ ] check on  safari
- [ ] check on  edge
- [ ] check on  ie 10
- [ ] check on  ie 11
- [ ] check on  android

catherine-jones:
ie 10 and 11 need to test on staging

catherine-jones:
testing completed

elisechant:
@catherine-jones we need to test at least with Voice Over on Mac.

app/views/dashboards/index.html.erb
```diff
@@ -60,14 +60,14 @@
       </div>
     </div>
     <div class="dashboards-intro-sec hide-on-desktop-tablet">
-      <div class="media">
+      <h2 class="media">
```
elisechant: This doesnt look right to me. Also, it's not a heading, please revert.

client/src/dashboard/styles/dashboard/_card.scss
```diff
@@ -17,9 +17,14 @@
   padding: $gutter $gutter $gutter 25px;
   color: $black;
 
-  h2 {
+  h3 {
```
elisechant: rather than do this, can we apply a h2 class on the element instead? like `<div class="h2">`

client/src/dashboard/styles/dashboard/_dashboard.scss
```diff
@@ -20,14 +20,21 @@
       @include span-columns(12);
     }
   }
+
+  h2 {
```
elisechant: and here

client/src/dashboard/styles/dashboard/_dashboard.scss
```diff
@@ -37,11 +44,18 @@
...
     }
+
+    .media {
+      height: auto;
+      float: left;
+    }
   }
-  @media(min-width: 768px) {
+
+  @media(min-width: 769px) {
```
elisechant: i'm not comfortable seeing `max-width:768px` and `min-width:768px`, it creates a 1px dead zone.

client/src/dashboard/styles/dashboard/_card.scss
catherine-jones: I don't think we need an additional class here because we are targeting the h3 inside the card class.

elisechant: my point is that we're redeclaring font styles. please use the .h2 or .h3 class on the semantic element to achieve the same result without adding more style rules

client/src/dashboard/styles/dashboard/_dashboard.scss
catherine-jones: @elisechant sorry I don't understand your comment here. I am avoiding using max-width and min-width at the same pixel value

client/src/dashboard/styles/dashboard/_card.scss
elisechant: another alternative:

```
%h2 { // styles ... }

.the-component h3 { @extend %h2; } 
```

the feedback is mainly about avoiding code duplication

catherine-jones: we aren't actually duplicating any code here because is base we are only declaring 
```
h3 {
  font-size: 1.25 * $fallback;
  font-size: 1.25rem;
  font-weight: 600;
}
```

PR #285, opened 2016-10-18
qa-fire that might work

tobyhede:


tobyhede:
Conflict in package.json

tobyhede:
Hmmm ... how to fork a fork and merge back ...

webpack.config.babel.js
```diff
@@ -5,7 +5,7 @@ import autoprefixer from 'autoprefixer';
 
 import * as CONFIG from './client/config/_config';
 const projectName = require('./package').name;
-const DEBUG = !process.argv.includes('--release');
+const DEBUG = !process.argv.indexOf('--release') !== -1;
```
elisechant: this needs rewrite.

package.json
```diff
@@ -143,7 +143,7 @@
     ]
   },
   "engines": {
-    "node": "6.3.0"
+    "node": "6.7.0"
```
elisechant: seems important to flag. why are we changing the node version?

webpack.config.babel.js
elisechant: @maxious what are you trying to achieve here?

package.json
tobyhede: @elisechant The buildpack no longer supports 6.3

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "dto-dashboard",
   "version": "1.0.0-beta",
   "description": "DTO Dashboard (beta)",
+  "preferGlobal": true,
```
tobyhede: We're building assets on CF, which means the CF container needs access to all the installed libs

PR #287, opened 2016-10-20
Hotfix/media queries

catherine-jones:


tobyhede:
So this refactors out some pixel values into variables?
Seems ok to me.

tobyhede:
Dont think this counts as a hotfix, btw. A hotfix is something you intend to push to production immediately. 

Its a refactor, maybe?

catherine-jones:
@tobyhede I do want to push it to prod straight away if that's ok?

client/src/dashboard/styles/_base.scss
```diff
@@ -93,19 +93,19 @@ h4 {
 }
 
 .hide-on-desktop {
-  @media (min-width: 769px) {
+  @media (min-width: $min-width-screen-md) {
```
elisechant: we need to review this implementation since introducing Bootstrap's Grid system

elisechant:
@catherine-jones why close?

PR #302, opened 2016-10-21
batch grid fixes

elisechant:
* implement Bootstrap


elisechant:
cancelled because already merged.

PR #309, opened 2016-10-23
PD-691 tests; npm hygiene; fix prod build bug;

elisechant:
* data focused tests
* simplify run scripts
* fix npm run build from qa-fire edits
* remove 2 unused deps


elisechant:
@catherine-jones @tobyhede npm run scripts change now to `npm run build` and `npm run dev` which is convention, no more `npm run webpack:dev`, not sure why we didn't make this change when we dropped Gulp.

PR #311, opened 2016-10-23
PD-695 Two-factor auth 

micapam:
Putting this up as a WIP PR, hoping QA Fire will build an instance for testing :)

I have a couple of questions on the design I need to check but this now broadly follows the UX Phil has laid out.


app/controllers/users/shared_secrets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    unless secret.present?
+      redirect_to new_users_shared_secret_path
+    end
+
+    if authenticate_code secret, params[:code]
+      current_user.update_attribute :otp_secret_key, secret
+      session.delete :setup_totp
+      redirect_to done_users_shared_secrets_path
+    else
+      @form_error = 'Your code didn\'t work. Please try again.'
```
ltankey: Whilst we are redirecting to `:new` and generating a new TOTP secret every time, we should probably still increment an attempt counter.  If they're chewing through 1000s of attempts, it's possible they're hunting for a collision.

app/controllers/users/shared_secrets_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  private
+
+  def authenticate_code(key, code)
+    code.present? && key.present? && ROTP::TOTP.new(key).now == code
+  end
+
+  def secret(value = nil)
+    @secret ||= session[:setup_totp] ||= value
```
ltankey: Does `session[:setup_totp]` every get cycled?  If not, we need an attempt counter.

app/views/users/shared_secrets/index.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="sign-in">
+  <h1>To create an account, you need to set up a 2-step verification with
+    Google Authenticator.</h1>
+  <h2>Step 1 of 2</h2>
+  <p>Download the <strong>Google Authenticator</strong> app from the App Store
+    or Google Play, and then select <strong>Set Up Account</strong>.</p>
+  <div class="qrcode">
+    <p><strong>Scan the barcode directly into your Authenticator
+      app.</strong></p>
+    <img src="/users/shared_secrets/code.png" />
```
ltankey: Would this ever get cached anywhere upstream? One option is to deliver the image directly with the content, but that may not be necessary if you have assurance it isn't cached.

PR #313, opened 2016-10-24
remove query for widget height

catherine-jones:


catherine-jones:
having issues testing locally on IE9 need to deploy to staging

PR #320, opened 2016-10-24
Fix - npm deps issues; move from npm to yarn;

elisechant:


elisechant:
@catherine-jones @micapam @tobyhede moving to Yarn instead of npm. Updated the readme with deets, but if you install yarn https://yarnpkg.com/en/docs/install everything should be the same as it was npm, instead of `npm install`, `yarn install` and instead of `npm run build`, `yarn run build`

PR #335, opened 2016-10-26
Very basic account status thing in the header

micapam:
Just a thing to show who you are when you are signed in (and a link to sign out).


catherine-jones:
screen shot 2016-10-27 at 11 56 33 am

I think maybe just centre it so it's more align with the performance dashboard part of the logo otheriwse all good

micapam:
Agree with your comment @catherine-jones - though this is just quick & dirty version, good one is in the works :)

PR #340, opened 2016-10-28
Fix for crash in prod env

micapam:
This is ugly as - checking for Rails env in a view :(
However I think until we have feature flags, ugliness is wearable. 
I just want this merged in quick so we don't forget & cause same error on next prod deploy.


elisechant:
🚢

elisechant:
@micapam maybe we make a task in JIRA or no point to bother?

micapam:
@elisechant I reckon a task in Jira for feature flags would be the one to make ... I can do it ? (in product backlog)

elisechant:
either or, just record the action if action should be taken, if not whatevs

PR #344, opened 2016-11-02
Fix - fix circle and update npm to yarn

elisechant:
PD-731
and fix circle which is broken for updating new npm deps


circle.yml
```diff
@@ -11,12 +11,16 @@ machine:
 
 dependencies:
   pre:
-    - npm install -g npm
-    - npm run build
+    - npm install -g yarn
+    - yarn install
+    - yarn build
```
elisechant: the node machine does a whole bunch of defaults that are non-desirable here like `npm install` - but don't think you can pick and choose defaults unfortunately. There's a redundant `npm install` that happens after assets are built. 

screen shot 2016-11-03 at 9 06 06 am

circle.yml
```diff
@@ -11,12 +11,16 @@ machine:
 
 dependencies:
   pre:
-    - npm install -g npm
-    - npm run build
+    - npm install -g yarn
+    - yarn install
```
elisechant: also, for some reason we are missing `npm install` before build which is the root issue.

elisechant:
@joahua summary is, preferGlobal was a big dirty hack that came through with QA Fire to "speed up" circle. it means all the npm deps were installed a single time in the global environment at which point any subsequent deps are not installed and then it breaks. Fix is to restore the `npm install` and remove preferGlobal.

micapam:
BTW I am not @joahua I am @micapam ;)

PR #345, opened 2016-11-02
Feature - sprint 41 editor iteration

elisechant:


app/admin/dashboard.rb
```diff
@@ -19,6 +20,7 @@
 
   form do |f|
     f.inputs "Dashboard" do
+      f.input :id
```
micapam: Why would an admin need to change the primary key??

client/src/_common/scripts/_vendor/gov-au-ui-kit/constants/date-time.js
```diff
@@ -1 +0,0 @@
-export const ISO_LONG_DATE = 'dddd D MMMM, YYYY';
```
micapam: Could this be causing that mismatch in the JSON date format? (from one of your messages in #dashboard-tech Slack channel @elisechant )

client/src/_common/styles/_variables.scss
```diff
@@ -1,13 +1,36 @@
-$dashboard-variables: (
+$dashboard-constants: (
+  // color palette
+  COLOR_CERULEAN: #0579c7,
```
micapam: I like this word so I looked it up & it's a gorgeous colour! Almost as nice as Yves Klein blue.

client/src/editor/scripts/components/forms/updateKpiDatagroup.js
```diff
@@ -0,0 +1,127 @@
...
+let UpdateTimeSeriesDatagroup = ({
+  isEditing, isSubmitting, onCancelSuccess, formModel,
+  ...rfProps
+}) => {
+
+  const { error, handleSubmit, pristine, valid } = rfProps;
+
+  return (
+    <form onSubmit={(e) => e.preventDefault()}>
+
```
micapam: Is a KPI always four?

app/admin/dashboard.rb
tobyhede: Yeah, should never edit the ID - it is generated by the database on create

elisechant: i want to be able to sort by id, using admin like a database

elisechant: oh wait. you're right. ok changing.

client/src/_common/scripts/_vendor/gov-au-ui-kit/constants/date-time.js
elisechant: nope. not sure about that i think it has to do with GMT.

client/src/editor/scripts/components/forms/updateKpiDatagroup.js
elisechant: technically yes. but we render the fields dynamically so if not then it would change to 3 say

app/admin/dashboard.rb
tobyhede: We can still give you the canonical ID if you need to sort.

PR #346, opened 2016-11-02
Disable 2FA is development mode; fix after sign in path

micapam:
Disable two-factor for local development; 
After sign in, go to editor (I made this change before, somehow it got lost :S )


elisechant:
looks like code... 🚢

app/controllers/application_controller.rb
```diff
@@ -13,7 +13,7 @@ def store_current_location
   end
 
   def after_sign_in_path_for(_resource)
-    session[:user_return_to] || editor_path
+    editor_path
```
elisechant: 👍  yes!

PR #350, opened 2016-11-04
Prevent 2FA reset if already set up

micapam:
Protects controller methods from `shared_secrets` if shared secret (2FA) has already been set up.

This prevents an unauthorised user who has somehow found a user's username & password from resetting 2FA. The bug fixed in this PR rendered 2FA moot.


tobyhede:
LGTM

PR #351, opened 2016-11-04
Fix 'generate a new code' and 'try again' in 2FA flow

micapam:
Links to DELETE methods require jquery-ujs which is overkill. 
Just switched out the links & used `button_for`... we can style the buttons to look like links if we want (thoughts?)


tobyhede:
I thought you could add a method to a link and rails would handle ... 
might have changed

micapam:
You can. But you need jquery-ujs for that (which we don't have). I guess we could include it as a node module?

tobyhede:
Nah, this is fine. Not worth adding the extra KB

tobyhede:
LGTM

PR #353, opened 2016-11-06
PD-850 JSON data is inconsistent

micapam:
This fixes the problem described in PD-850: inconsistency (in the editor JSON) between datasets belonging directly to the user, and datasets indirectly related to the user via widgets.

But I'm not sure this is the best way to do it. It's a hack. Do we need to rationalise / normalise the relationship between users, datasets and widgets? There seem to be too many linkages & thus the data can get confused.


elisechant:
@micapam yeah it gets very complex on client side around all of these linkages as well. Let's go over issues together with @tobyhede

elisechant:
i'm also creating `widgetDatasets` etc.

tobyhede:
I've updated this to make sure that a user is given access to all the datasets in a dashboard. 
This resolves the immediate concern

app/models/user.rb
```diff
@@ -64,6 +64,14 @@ def to_s
     email
   end
 
+  def add_datasets_to_user(dashboard)
+    datasets << dashboard.datasets.all
+  end
+
```
micapam: (Then remove method will screw up data)

micapam:
'Approve' is disabled for me (I guess because I started this PR? Even though it's reassigned? Get it together github...) but LGTM

PR #358, opened 2016-11-08
Generate api token on successful page render

tobyhede:
At the moment, the token is generated when the user has been authenticated by devise.

Because of the way the 2fa flow works, the check only occurs when a controller attempts to validate the user via `authenticate_user`.  

This means that a token is generated for the user, even though they are not given access to it - the page redirects before rendering, but it is a possible vector. 

This change ties the token generation to a successful request of the editor page.


app/models/user.rb
```diff
@@ -65,7 +61,7 @@ def to_s
   end
 
   def need_two_factor_authentication?(_request)
-    !Rails.env.development? && !Rails.env.staging?
```
micapam: Probably want to back this out ;)

tobyhede:
Rebased and caught your feature flag changes

PR #368, opened 2016-11-09
Updated invitation email content

micapam:
Content updated as per Nigel's comments. Handles invitation to multiple dashboards. 

image


tobyhede:
:shipit:

PR #369, opened 2016-11-10
Stop deploying JSON seed data on production

micapam:
Once we are able to manage the data via the editor, we need to stop automatic population on every deployment from the JSON.


bin/cf-start.sh
```diff
@@ -2,7 +2,4 @@
 
 set -xeu
 
-bundle exec rake cf:on_first_instance db:schema:load import:data
-bundle exec rake cf:on_first_instance bootstrap:admin_user
```
tobyhede: Actually ... we've never set this up on prod because waiting for 2FA.

PR #370, opened 2016-11-10
Idea for datapoints

micapam:
Sketch of how datapoints structure would look if we removed timestamp & just indexed them with start timestamp & period on the dataset. 

Advantage to this is that we don't have to worry about complex validation rules and pulling out / pushing in synchronic slices of datapoints across datasets would become trivial. 

Less drastic way could keep `ts` column but make it nullable (i.e. only for datasets that aren't going to be edited via the editor, e.g. external APIs). In this case `ts` getter could be like: 

```
def ts
  self[:ts] || dataset.datapoint_ts(self)
end
```

### DIFF
--- a/app/models/datapoint.rb
+++ b/app/models/datapoint.rb
@@ -1,12 +1,16 @@
 class Datapoint < ApplicationRecord
   belongs_to :dataset
 
-  validates :ts, :presence => true
+  validates :idx, presence: true, uniqueness: { scope: :dataset_id }
 
   validates :value, :numericality => true, :allow_nil => true
 
-  scope :by_time, -> { order(:ts => 'ASC') }
-  scope :by_time_desc, -> { order(:ts => 'DESC') }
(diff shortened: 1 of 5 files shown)


tobyhede:
I'd be a bit nervous about not keeping the timeseries data itself - removing the ts altogether would mean we would only ever be able to store data that fell within a predetermined time period (day, week, month). 

But, I think the overall idea would cleanup some of the front-end complexity. 
The editor only cares about a group of values for the period. The back-end can then do the heavy lifting.

tobyhede:
Ah, and remember that the charts still need the data in non grouped format.

micapam:
@tobyhede You could maybe indicate that a dataset has 'free' datapoints by a null value in `period` or even `period = 'free'`

tobyhede:
I think in most cases for artisinal data, its a reporting period.

PR #375, opened 2016-11-11
Bug/pd 866 dataset missing json

micapam:
Datasets were missing from JSON. This was because the update action on the activeadmin controller was wiping out any datasets not manually selected. 

I've moved the logic to ensure dashboards' datasets are included in the user from the model to the activeadmin controller.


tobyhede:
Good catch

app/models/user.rb
```diff
@@ -60,10 +60,6 @@ def to_s
     email
   end
 
-  def add_datasets_to_user(dashboard)
```
micapam: Ah thanks, I'll sort that out.

PR #376, opened 2016-11-11
Link to invitation in admin/users#index instead of 'New' action

micapam:
Best to use the invitation flow rather than 'new' action in ActiveAdmin which sends some other weird email to the user... I've linked to the invitation form from the toolbar in the users section of the admin, and removed the 'New' button.


tobyhede:
LGTM

PR #381, opened 2016-11-14
Remove account info from public-facing header

micapam:


tobyhede:
:+1:

tobyhede:
We could just disable this for production ... its kinda handy to have the link when testing.

micapam:
It's already disabled on prod though? https://github.com/AusDTO/dto-dashboard/blob/master/app/views/layouts/_account.html.erb

PR #385, opened 2016-11-14
Periodic validations

micapam:
Added a `period` attribute on Dataset. Valid values so far: `free`, `month` (defaults to `month`).

If `free`, datapoints for that dataset can be added with any `ts` value (as now).

If `month`, validation on the datapoints will ensure that any datapoint added is not in the same month as any existing datapoint. 

We could extend this in obvious ways for week, day, hour etc.


app/models/datapoint.rb
```diff
@@ -5,11 +5,27 @@ class Datapoint < ApplicationRecord
...
+  scope :of_year, -> (year) { where('EXTRACT(YEAR FROM ts) = ?', year) }
 
   def label
     ts.strftime("%Y-%m")
   end
 
+  private
+
+  def check_frequency
+    if dataset&.period.present?
```
tobyhede: That is nifty

app/models/dataset.rb
```diff
@@ -2,6 +2,8 @@ class Dataset < ApplicationRecord
   include Measurable
   include Nameable
 
+  PERIODS = %w(month week day)
```
tobyhede: Is it worth setting a default?

micapam: Currently default is nil (which is a valid value) - do you think it should be `month`? For instance - would all our current datasets be monthly?

micapam:
Good point re. the import data - I will do that - or set a default as you mentioned (same result).

app/models/dataset.rb
tobyhede: Everything is monthly at the moment. 

I think. 

Pretty sure. 

cc @elisechant ?

elisechant: yes monthly or no period

micapam:
can haz merge ?

tobyhede:
:shipit:

PR #390, opened 2016-11-14
Override default dependencies

tobyhede:


circle.yml
```diff
@@ -11,9 +11,19 @@ machine:
 
 dependencies:
   pre:
+    - npm ls -gp --depth=0 | awk -F/ '/node_modules/ && !/\/npm$/ {print $NF}' | xargs npm -g rm
+    - npm -g uninstall strip-ansi which
```
elisechant: awesome and terrifying! think we can drop these 2 lines :

```
+    - npm -g uninstall strip-ansi which
 +    - rm -rf node_modules
```

circle.yml
```diff
@@ -11,9 +11,19 @@ machine:
...
+    - rm -rf node_modules
+    - npm cache clean
     - npm install -g yarn
+    - yarn cache clean
```
elisechant: is this a thing?

package.json
```diff
@@ -110,7 +111,8 @@
     "watchify": "^3.7.0",
     "webpack": "^1.13.2",
     "webpack-dev-server": "^1.15.0",
-    "whatwg-fetch": "^1.0.0"
+    "whatwg-fetch": "^1.0.0",
+    "which": "^1.2.12"
```
elisechant: need to remove `which` and `strip-ansi` from here

elisechant:
LGTM

PR #395, opened 2016-11-15
PD-886 mark the dataset as updated when datapoiunt created

tobyhede:


tobyhede:
@micapam Is there a better way to test that a field was updated?

micapam:
Ohhhh OK - actually changing `updated_at` on the dataset - I thought you'd just decorate the attribute?

micapam:
Something like 

```
def updated_at
  [datapoints.order(updated_at: :desc).first.updated_at, object.updated_at].max
end
```

tobyhede:
We could do that too. 
Only issue with that is that the dataset can be edited too, its update value isn't necessarily tied to data.

tobyhede:
Which makes me think ... this is a new field for "data_updated_at" as distinct from updating a name or label

micapam:
@tobyhede Yeah a separate field which is not affected by the dataset's own `updated_at` sounds like a good idea. Which could go on the model, no need for a decorator... maybe something like: 

```
def data_updated_at
  datapoints.by_time_desc.first.updated_at
end
```

... ?

tobyhede:
Jinx!

micapam:
Ah hahah! But yours is also null-safe :)

tobyhede:
Using the cool idiom i learned off you yesterday

PR #398, opened 2016-11-15
Fix dupes, option 1 :)

micapam:
This is a basic way of fixing the duplicates mentioned in PD-871 - just removing dupes via a migration, adding a unique constraint on the HABTM joining tables, and fixing a bug in a rake task that was causing dupes to be added during import.

Note that #399 is another PR with option 2 (which I slightly prefer...) to solve this same problem.


micapam:
We went with option 2 so closing this one.

PR #399, opened 2016-11-15
Fix dupes, option 2

micapam:
Fixing for the duplicates mentioned in PD-871 - removes the HABTM association and replaces it with a Permission class. (All associations are migrated and de-duped in the process.) Also fixes a bug in a rake task that was causing dupes to be added during import.

This is my preferred option as it's more extensible than option 1: #398 (also this gets rid of HABTM)


tobyhede:
I think this is definitely the way to handle this.

PR #400, opened 2016-11-15
Stop the periodic validations breaking all the specs!

micapam:
When I introduced periodic validations I forgot to set the dataset factory to generate 'free' by default.


tobyhede:
👍

PR #401, opened 2016-11-16
Install yarn for circle

tobyhede:


elisechant:
💯

elisechant:
:shipit:

elisechant:
side not we can drop the `"` around the command on line 9

elisechant:
@tobyhede why did this PR fail earlier? did you "Rebuild without cache" in circle?

tobyhede:
I had done a rebuild without cache, but then it failed because cache. It would be picking up the cache from another build - they are shared. So once this is in, we do the global clean cache again, and then this fix should ensure we dont have the sme npm problems ever again

elisechant:
oh cool. makes sense.

PR #402, opened 2016-11-16
widget metadata pass 2

elisechant:


client/src/editor/scripts/data/widgetMetadata.js
```diff
@@ -63,20 +63,20 @@ const metadata = {
     '%': {
       label: '',
       widget_help: `This is represented as a single-line chart. Enter your data in the fields below.`,
-      widget_form_help: '',
+      widget_form_help: `This figure is a percentage and can't be more than 100. You can include up to 2 decimal places — for example, 32.25.`,
```
elisechant: suggest we need "these figures added together can't be more than 100"

elisechant:
Suggest we need:
* Fact - widget_form_help
* Sparkline - form_help && widget_form_help

client/src/editor/scripts/data/widgetMetadata.js
```diff
@@ -22,13 +22,13 @@ const metadata = {
   'bar': {
     'n': {
       label: '',
-      widget_help: 'This is represented as a single-point bar chart. Enter your data in the fields below.',
```
elisechant: can't use "single-" for bar or line because there's no good way of identifying that dynamically at this point in time without changing chart type aliases

PR #407, opened 2016-11-17
Redirect user when the API is unauthorized

tobyhede:
Redirects to /editor, which will refresh the API token if the token has expired, but the users session is invalid. 

If the session itself is invalid, /editor will bounce the user to sign-in


tobyhede:
@elisechant is this ok?

client/src/editor/scripts/actions/datagroupset.js
```diff
@@ -30,7 +30,7 @@ export const createDatagroupset = formData => {
         }
       })
         .then(response => {
-          if (status >= 200 && status < 300) {
```
tobyhede: The service endpoint should always return 201 on success.

app/controllers/api/v1/api_controller.rb
```diff
@@ -44,6 +44,6 @@ def authenticate_token
 
   def render_unauthorized
     self.headers['WWW-Authenticate'] = 'Token realm="Application"'
-    render json: 'Bad credentials', status: :unauthorized
+    render :json => { :code => '401', message: 'Method not Found' }, :status => :unauthorized
```
elisechant: this is great

client/src/editor/scripts/actions/datagroupset.js
```diff
@@ -30,7 +30,7 @@ export const createDatagroupset = formData => {
...
+          if (status != 201) {
```
elisechant: this code is not reachable because the former code is not reachable, it needs to be `response.status`. I have changed this is my today branch, so can ignore.

client/src/editor/scripts/actions/datagroupset.js
```diff
@@ -55,7 +55,9 @@ export const createDatagroupset = formData => {
         return responses;
       })
       .catch(err => {
-        debugger;
+        if (err.message == '401') {
+          window.location.replace("/editor");
```
elisechant: I also handled this in my branch today, so can ignore. At any rate, probably the location should be more like `sign-out` instead.

tobyhede:
@elisechant Should I close this one?

elisechant:
@tobyhede can we keep changes to app/controllers/api/v1/api_controller.rb

tobyhede:
Removed other changes, keeping the api json change. Will merge on green

PR #412, opened 2016-11-17
Add widget.data_updated_at

tobyhede:
Once editor is live, the last_updated_at field can be deprecated.

We can’t change last_updated_at until the editor has rolled into prod and we’ve stopped deleting the db on deploy


micapam:
LGTM

PR #421, opened 2016-11-20
api management improvement; kick off if not auth;

elisechant:
* deprecate config reducer; 
* adds basic api wrapper for thunk; 
* kick out if no token;


client/src/editor/scripts/api.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @param route - API endpoint
+ * @param token - API session token
+ * @param options - options to pass to Fetch
+ * @return {Function<Promise>}
+ */
+export default function api(route, token, options={}) {
+
+  if (!token) {
```
elisechant: @tobyhede i create an abstraction to handle api. Here I can kick out if there's no token.

client/src/editor/scripts/api.js
```diff
@@ -0,0 +1,45 @@
...
+    headers: {
+      'Accept': 'application/json',
+      'Content-Type': 'application/json',
+      'Authorization': `Token ${token}`
+    }
+  }, options))
+    .then(response => {
+      if (response.status === 401) {
+        throw {
+          message: `Unauthorised. If you think you should be authorised try logging out and then logging in again.`,
```
elisechant: @tobyhede here i can check for 401. Was thinking to display message might be better than kicking out for this case. Thought ?

client/src/editor/scripts/api.js
```diff
@@ -0,0 +1,45 @@
...
+    credentials: 'same-origin',
+    headers: {
+      'Accept': 'application/json',
+      'Content-Type': 'application/json',
+      'Authorization': `Token ${token}`
+    }
+  }, options))
+    .then(response => {
+      if (response.status === 401) {
+        // rather than kick out, pass them a message
```
elisechant: @tobyhede thoughts?

tobyhede: That's probably a good idea. If they refresh, it will either bounce them or renew the token.

PR #429, opened 2016-11-22
Simple previews

micapam:
Loads a preview inline from the server and pushes it onto the page. 

Ugly as all hell but working - except that chart styling isn't happening (numbers show that the preview is working) ...


elisechant:
So as discussed, we should make this an iframe. And we'll need to rework some of the React parts to be more React-land. But WIP.

micapam:
Ready for reviews. The JS code is nasty. But - once the hero preview is fixed - could work as an MVP for user testing.

app/decorators/previewing_dataset_decorator.rb
```diff
@@ -0,0 +1,31 @@
+class PreviewingDatasetDecorator < Draper::Decorator
```
micapam: Good point

app/views/dashboards/export.csv.shaper
```diff
@@ -0,0 +1,15 @@
+csv.headers :dataset_name, :units, :time_stamp, :label, :value
```
tobyhede: Not seen this gem before, nice one.

tobyhede:
Looks good, I'd add a spec for the decorator, even as mvp

micapam:
Spec has decorator now @tobyhede

tobyhede:
LGTM 👍

elisechant:
cool, but we're going to need to do a version 2 to improve some of the front end codes

elisechant:
created https://govausites.atlassian.net/browse/PD-930

PR #432, opened 2016-11-22
alternative - dont include images in asset pipeline

elisechant:
cc @micapam but means we cant use rails image helpers


tobyhede:
What's the rationale?

elisechant:
yeah i dont like this idea. i like the rails pipeline managing cached assets

PR #433, opened 2016-11-22
webpack tweaks and update deprecation notices

elisechant:
* update deprecation for redux dev tools
* adds image compression
* adds source maps by default
* remove render delay for editor and login


tobyhede:
nice 👍

PR #440, opened 2016-11-23
Feature - icon loader 

elisechant:


elisechant:
screen shot 2016-11-23 at 2 54 29 pm

PR #441, opened 2016-11-23
Ensure user is 2 factored before granting access to flipper

tobyhede:


config/initializers/flipper.rb
```diff
@@ -6,7 +6,11 @@
 class CanAccessFlipperUI
   def self.matches?(request)
     current_user = request.env['warden'].user
-    current_user.present? && current_user.respond_to?(:admin?) && current_user.admin?
+    CanAccessFlipperUI.two_factor_authenticated?(request) && current_user.present? && current_user.respond_to?(:admin?) && current_user.admin?
+  end
+
+  def self.two_factor_authenticated?(request)
+    request.env['warden'].session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] != true
```
tobyhede: There might be a better way of doing this

tobyhede: I tested it and seemed ok, but good idea anyway

PR #449, opened 2016-11-28
fix - resize text if fact widget overflows fixed height container on desktop

elisechant:
* patch fix, nothing glamorous, just to fix the bug for if text is longer than allocated space


client/src/dashboard/legacy/Widgets/FactWidget.js
```diff
@@ -1,16 +1,47 @@
...
+    if ($(window).innerWidth() > 992) { // sass: $screen-md
+
+      // resize the font-size if it overflows the container
+      let $parent = this.$el.find('.fact__inner');
+      let $child = $parent.find('p');
+
+      let size;
+      let minSize = 10;
+      while ($child.height() > $parent.height()) {
+        size = (parseInt($child.css('font-size')) - 1);
```
elisechant: use `Number` not `parseInt`

tobyhede:
👍 React on the Dashboard!!

PR #455, opened 2016-11-30
Refactor import

micapam:
## Background

Historically, the data for the Performance Dashboard project has been stored in JSON files in the codebase, which are then used (via the rake task import:data) to populate the database on each deployment. This is still the case; but we have set up an admin system to allow authorised users to edit their data. Once we start using that admin system, the importing script is no longer going to be used to import data every deploy, as the database must become the source of truth rather than JSON files.

However, we still have a need for imports, so that new data can be set up as we onboard organisations to the Performance Dashboard. The Dashboard team is familiar with the JSON structure, so as an MVP, we are going to allow import of JSON data via an upload screen on the admin.

## Changes

This PR just includes the refactoring of the import functionality so it's no longer just sitting in a rake task (with a view to using the new OrganisationImporter class in the admin upload as described above). Also, a basic spec has been included around the import function. A later PR will implement the actual admin functionality.


spec/lib/organisation_importer_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      specify { expect(dashboard.name).to eq(
+        'Australian Citizenship Appointment Booking Service') }
+
+      specify { expect(dashboard.widgets.count).to eq 9 }
+
+      specify { expect(dataset.datapoints.first.value).to eq 99 }
+
+      specify { expect(dataset.widgets.count).to eq 2 }
```
micapam: Good idea. I'll do that.

lib/tasks/import.rake
```diff
@@ -14,114 +15,14 @@ namespace :import do
...
-      file = File.read("lib/data/#{name}-data.json")
-      data = JSON.parse(file)
-
-      file = File.read("lib/data/#{name}-definition.json")
-      definition = JSON.parse(file)
-
+      data_json = File.read("lib/data/#{name}-data.json")
+      data = JSON.parse(data_json)
+      definition_json = File.read("lib/data/#{name}-definition.json")
       organisation = Organisation.find_or_create_by!(:name => data['agency'], :url => data['url'])
```
micapam: In context of this PR, yeah... but in the context of using the importer in the admin (next step) it's easier to feed the organisation to the constructor.

tessereth: Cool. I just wanted to check 😄

PR #456, opened 2016-12-01
Importing JSON data via admin 

micapam:
Following on from the refactor of the import code (see #455), this PR allows authorised admin users to import dashboard data, using the same JSON format that has up to this point been used to populate the database via a rake task during deployment. 

I have also dug up the JSON schemas that were used a while ago to validate the original import data (they were stored in [another repo](github.com/AusDTO/dto-dashboard-data)). I've updated them with some small changes to conform with ways in which the JSON structure has changed since they were written, and added them to codebase - they are now used to validate the data prior to import.


app/views/admin/organisations/import_form.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= semantic_form_for @organisation, url: { action: :import, method: :patch } do |f| %>
+
+  <h2>Warning</h2>
+  <p>Importing data for the organisation <strong><%= @organisation.name %></strong>
+    will overwrite all its existing data with the JSON data you upload below.
+    Any edits made to the existing data will be lost.</p>
+  <p><strong>YOU HAVE BEEN WARNED.</strong></p>
+  <p>&nbsp;</p>
```
tessereth: I assume you have a good layout reason for this. But it's still sad.

spec/controllers/admin/organisations_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    it 'calls the importer' do
+      # Bypass auth
+      allow(subject).to receive(:authenticate_user!).and_return true
+      allow(subject).to receive(:current_user).and_return user
+
+      # Does the import
+      expect(OrganisationImporter).to receive(:new).and_call_original
+
+      # Clears the cache
+      # expect(subject).to receive(:invalidate_dashboards).and_call_original
```
tessereth: Does this need to be checked or can it be removed?

app/views/admin/organisations/import_form.html.erb
micapam: It is :) I guess the reason is that is an MVP for getting a dashboard off the ground - basically doing what we used to do with rake tasks, but within the admin. The end goal is to have a much nicer way of setting up dashboard data within the editor (and let organisations bootstrap their own dashboards rather than just the Dashboard team having to do it). But that's a lot of work. 

I should have mentioned - the admin area is for DTA access only.

PR #461, opened 2016-12-02
update october data

mibraheim:
for all datasets


lib/data/imports-data.json
```diff
@@ -14,7 +14,9 @@
         "units":        "%",
         "name":         "User satisfaction",
         "note":         "We do not have this data at the moment because we are....",
```
micapam: Maybe this not is no longer relevant?

mibraheim: updating it in the definition schema now :)

PR #469, opened 2016-12-09
Dashboard notes refactor

micapam:
## Background

Since the original design of the Performance Dashboard, the `notes` field (for contextual information about a dashboard) has been stored in a JSON attribute as a snippet of raw HTML. This HTML is then stored in the database, and rendered directly onto the `dashboards#show` page. My guess is that this was done to allow flexibility in terms of the HTML content that could be shown on a dashboard.

## Problem

Although currently there is no access to the editor or admin functionality on the production servers, it’s not good security practice to be storing or rendering raw, unsanitised HTML from the database. Furthermore, we are soon to enable access to the editor and admin, which will make this a more serious security issue.

## Solution

Since all notes on all extant dashboards follow the same structure of HTML tags, the flexibility point is moot. Even bracketing security concerns, it’s unwise to allow even admin users to edit HTML that will then be directly served onto the public-facing website. I have therefore made the `notes` field serialised, and defined a simple JSON structure for contextual notes that supports the HTML currently output, for which this PR is a transparent change.

I have also updated the definition JSON files for each dashboard (changing `notes` nodes from `string` to a structured type), added a view partial that renders the JSON notes as HTML, modified the admin to ensure JSON is properly displayed and updated, and updated the JSON schema for import validation.

*n.b.* Once merged, this PR should be released to production separately from, and before, the deprecation of the JSON seed data (#369). This is because the dashboards need to be repopulated one final time using the updated definition files.


app/views/dashboards/show.html.erb
```diff
@@ -80,10 +80,10 @@
                     <h4>Dashboard Information</h4>
                   </div>
                   <div id="dashboard-notes" class="widget__body">
-                    <% if dashboard.show_notes? %>
-                        <%= raw dashboard.notes %>
+                    <% if dashboard.show_notes? && dashboard.notes.any? %>
```
tessereth: Would it make more semantic sense for `dashboard.show_notes?` to check if there are any?

lib/data/industry-definition.json
```diff
@@ -4,7 +4,24 @@
...
...faction measures the percentage of users who responded “yes” to finding the service useful in the feedback survey as compared to the total number of respondents. <dt> Cost per transaction <dd> We do not have this data at the moment because we are currently in the process of determining the best way to measure Cost per transaction for information services. <dt> Digital take-up <dd> Digital take-up measures the percentage of visitors to the page on business.gov.au, where the tool can be found, who go on to use it. The percentage is calculated by comparing the total number of users who access the public beta tool to the total number of unique visits to the page on business.gov.au. We are working on a more precise measure of take-up, comparing use of the tool to users accessing this information through our contact centre. We will update the Dashboard in the near future with this data. <dt> Completion rate <dd> Completion rate measures the percentage of users who successfully generate a result by using the public beta release of the tool. This percentage is calculated by dividing the total number of successfully completed interactions by the total number of started interactions. </dl>",
```
tessereth: This line should be removed

lib/data/mygov-definition.json
```diff
@@ -185,7 +185,23 @@
     }
   ],
   "url":    "https://my.gov.au",
-  "notes":  "<dl> <dt> Where is this information from? <dd> myGov & Australian Department of Human Services. <dt> What is myGov? <dd> myGov is a secure, fast and simple way to access Australian Government services online with one username and password. <dt> What is an \"Active Account\"? <dd> An account on myGov with at least 1 linked member service. <dt> What is a \"Sign-in\"? <dd> A log in to the myGov service. In a given month, multiple sign-ins by a single account will be counted multiple times. <dt> What is a member service? <dd> This is an Australian Government service available to users through the myGov account. </dl>",
+  "notes":  "<dl> <dt>  <dd>  <dt>  <dd>  <dt>  <dd>  <dt>  <dd>  <dt>  <dd>  </dl>",
```
tessereth: This line should be removed

spec/fixtures/valid-definition.json
```diff
@@ -95,8 +95,21 @@
...
+    { "title":  "User satisfaction",
+      "body":   "User satisfaction measures the percentage of users who responded \"satisfied\" and \"very satisfied\" with the appointment booking service using the feedback kiosks located in the Department’s citizenship areas." },
+    { "title":  "Cost per transaction",
+      "body":   "This data is not available." },
+    { "title":  "Digital take-up",
+      "body":   "Digital take-up measures the number of users who had an appointment booked using the new service as a percentage of all appointments that have been made Australia-wide." },
+    { "title":  "Completion rate",
+      "body":   "Completion rate measures the number of users who were able to use the digital service without having to contact the Department for assistance when rescheduling their appointment." }
+  ],
+  "url": "https://www.border.gov.au/",
```
tessereth: Just checking, did you mean to change this url?

lib/data/industry-definition.json
micapam: Indeed!

spec/fixtures/valid-definition.json
micapam: I think it was (elsewhere) changed in its source (`dibp-definition.json`) and when I copied it into this spec file to update the notes, URL was also changed here... doesn't make much difference either way though. URL is only ever used to populate a link.

micapam:
I added a test to the organisation importer spec to check how many notes are added... reckon it needs more @tessereth?

app/decorators/dashboard_decorator.rb
```diff
@@ -10,7 +10,7 @@ def show_kpis?
   end
 
   def show_notes?
-    notes.present?
+    notes.present? && notes.any?
```
tessereth: Hmmm. I seem to remember that `#present?` returns false for an empty array... \<goes and checks\>
```
pry(main)> [].present?
=> false
pry(main)> ['foo'].present?
=> true
```
I think that means the `#any?` call is redundant.

micapam: Good point :)

PR #470, opened 2016-12-11
Feature/pmnc dashboard

micapam:
Just a WIP pull request to share the PM&C data with the team.


elisechant:
screen shot 2016-12-19 at 10 47 07 am
 
added Salmon and Pink

elisechant:
@mibraheim I think we still need to come up with a solution for actually fitting the legend in the chart space.

micapam:
Are any of the legend bits missing @elisechant ? Or is it just that there is not enough margin at the bottom?

elisechant:
@micapam ahh.. not sure. I think just padding. actually @mibraheim we do need to think about this issue, not just the legend key length which we discuss yesterday

PR #473, opened 2016-12-16
tweak - changes urls from /sign-in to /login etc

elisechant:
I know this is what the default is for gov.au but it was annoying me and i dont think its the right convention


elisechant:
@micapam can you think of any other places this would need updating?

PR #474, opened 2016-12-16
remove uikit from login

elisechant:
* moves uikit stuff to Bootstrap like Editor


elisechant:
@micapam can you put an eye on the Rails syntax, not sure I know how to run some of the 2FA routes to verify AOK

elisechant:
@micapam i've redone this piece of work without removing the uikit (less risk for now), so we can close this PR. I am concerned though that there are links in the login area that go no where. I think we still need to put an eye on the functional behaviour of links in this part of the site.

PR #476, opened 2016-12-18
rwd header footer for Editor

elisechant:


elisechant:
a lighter alternative to https://github.com/AusDTO/dto-dashboard/pull/474 without remove uikit from login

PR #479, opened 2016-12-19
adds commitizen

elisechant:


elisechant:
@micapam thoughts?

elisechant:
from Slack: 
"pops a gui to help you write better commit messages, and then plugs in to git hooks to validate that the messages are in the expected shape"

screen shot 2016-12-19 at 9 33 32 am
screen shot 2016-12-19 at 9 32 38 am

micapam:
Looks cool but how about we use this too? https://www.npmjs.com/package/cz-jira-smart-commit

elisechant:
@micapam ohhhh! good idea

elisechant:
also, which commit types we would like, maybe they should marry to JIRA issue types, as in "task", "bug"

micapam:
Sounds good... 'story', 'task', 'bug' ?

PR #480, opened 2016-12-19
Remove sign in button from header on sign in, views refactoring

micapam:
Hides log in link from login page. Also refactored condition as we show stuff based on flipper feature flag, not Rails env.

Also, turns out that `_account.html.erb` partial wasn't being used anywhere so I ditched it.


app/views/partials/_header.html.erb
```diff
@@ -14,13 +14,15 @@
...
       <div class="govau-header__container__utils">
-        <% if !Rails.env.production? && current_user.present? -%>
-          <div class="user-email">
-            <%= link_to current_user.email, user_path %>
-          </div>
-          <%= link_to 'Sign out', destroy_user_session_path, class: "UIK-button btn sign-out" %>
-        <% else %>
-          <a href="/login" class="UIK-button btn sign-out">Sign in</a>
+        <% if show_auth? %>
+          <% if current_user.present? -%>
```
elisechant: what is `-` at `-%>` do?

micapam: Just removes trailing whitespace in the output: http://apidock.com/rails/v4.2.1/ActionView/Base

PR #481, opened 2016-12-19
Refactored helper methods and decluttered views

micapam:
Transparent refactors. Just DRYs up the views & renamed some helper functions to conform to Ruby standard (snake rather than camel case).


app/helpers/application_helper.rb
```diff
@@ -28,16 +28,26 @@ def hashAsset(public_relative_path)
     public_relative_path + "?" + Digest::MD5.file(File.join(Rails.public_path, public_relative_path)).hexdigest
   end
 
-  def getPublicImageSrc(relative_path)
+  def public_image_source(relative_path)
```
elisechant: good idea

app/helpers/application_helper.rb
```diff
@@ -28,16 +28,26 @@ def hashAsset(public_relative_path)
...
     hashAsset('/images/' + relative_path)
   end
 
-  def getPublicStylesheet(relative_path)
-    tag('link', href: hashAsset('/stylesheets/' + relative_path), rel: 'stylesheet')
+  def public_stylesheet(sheet)
+    # Not required for webpack-dev-server mode
+    unless development_server?
```
elisechant: I know it DRYs things up, but I felt that having that logic in the layout view kept the separation between the backend and frontend.

app/helpers/application_helper.rb
elisechant: what if we pass the extension as a second parameter so its alike the js and css ones like: `<%= public_image_source('editor-logo', 'svg') %>`

app/helpers/application_helper.rb
micapam: Agree with the principle of separation of concerns but this is what view helpers are for - to encapsulate view logic. View helpers are part of the Rails front end.

app/helpers/application_helper.rb
micapam: Great idea, I'll see which is most common so we can use as a default.

PR #482, opened 2016-12-19
show a message if editor fails in production

elisechant:


elisechant:
screen shot 2016-12-19 at 12 02 56 pm

client/src/editor/scripts/main.js
```diff
@@ -6,14 +6,28 @@ import { syncHistoryWithStore } from 'react-router-redux';
 
 import configureStore from './store/configureStore';
 import initialState from './store/initialState';
-import Root from './containers/root';
 
 
 const bootState = merge(initialState, window.__STATE__);
+delete window.__STATE__;
 const store = configureStore(bootState, hashHistory);
```
elisechant: probably should wrap the whole of the Redux booting in the try catch down here as well

PR #484, opened 2016-12-19
PD-971 - update phase banner like Marketplace

elisechant:


config/initializers/application_constants.rb
```diff
@@ -0,0 +1,5 @@
+# General application constants
+
+module Constants
```
elisechant: @micapam are you ok about naming and implementation here?

micapam: Maybe DashboardConstants would make the intent clearer?

elisechant: ApplicationConstants ? its used in login

PR #490, opened 2016-12-22
ci - fixing circle through error in build without remove cache

elisechant:


elisechant:
@micapam i think this fixes the circle issue, but don't really know

micapam:
If it builds OK, let's merge & we will see on the next rebase :)

PR #498, opened 2017-01-04
Data update for imports dashboard

micapam:


elisechant:
@micapam suspect this didn't get to test because tests were failing the build. Can we try again now? TLDR is that the patch fix for dot is to not render the chart is data is non sequential, test was testing case otherwise.

PR #499, opened 2017-01-04
css and icons for Rails Alerts

elisechant:


elisechant:
@micapam it would probably make sense to have the icons as client/src/_shared/scripts/components/_iconData.js as a json resource instead and then require and then iterate over it it in JS land as well Rails Land.  But yeah, taken the short cut here.

PR #504, opened 2017-01-05
fix(charts): make stroke dashes different line lengths

elisechant:
screen shot 2017-01-05 at 1 59 31 pm
screen shot 2017-01-05 at 1 58 21 pm


client/src/dashboard/legacy/Helpers/strokeDashes.js
```diff
@@ -0,0 +1,20 @@
...
+  '16,6',
+  '6,2',
+
+  '2,6',
+  '2,2',
+  '8,6',
+  '6,2,2,2',
+  '2,6,8,6',
+  '16,6',
+  '6,2'
```
bjgordon: I'm curious about the duplicate entries in the list... If the array needs to be longer than 7, should they all be unique?

elisechant: It's a hack, just in case they run out. yes you are right. Currently we only have 7 colours, although that is also not really an excuse.

elisechant: @bjgordon added notes in comments to explain https://github.com/AusDTO/dto-dashboard/pull/505/files. thanks mate.

bjgordon: @elisechant all good 👍

PR #516, opened 2017-01-06
Improve validation error

micapam:
Error message that was being surfaced was weird & unhelpful. This makes it make sense.


elisechant:
👍

PR #519, opened 2017-01-07
PD-994 reducer and helper tests

elisechant:


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.3'
+ruby '2.3.1'
```
elisechant: uncommit this

PR #532, opened 2017-01-10
feat(charts): new category grouping line chart type; adds imports data;

elisechant:


lib/data/imports-data.json
```diff
@@ -51,19 +51,35 @@
...
+      ]
+    },
+    {
+        "id":           "clearing-time-2016",
+        "recorded_at":  "2016-09-07T01:01:01.111Z",
+        "units":        "i",
+        "name":         "2016",
+        "note":         "This shows the average time taken to clear a permit using the Permit Verification Service.",
+        "data": [
+          { "label": "2015-05", "value": 83007},
```
elisechant: @micapam this is the "hack" - labels must correspond here see "clearing-time-2015" and "clearing-time-2016", labels both prefixed "2015-", otherwise the charts rendering along the x-plane and hidden from view.

elisechant:
screen shot 2017-01-11 at 8 37 44 am

elisechant:
need to do a final verification on data before pushing out

elisechant:
mariam [3:46 PM] 
@elisechant looks good! just a couple of things. Do you know why the 2016 line is dropping below the x-axis?  Also, the labels should just read May, Aug, Nov without the year value

PR #534, opened 2017-01-12
Test - editor root reducer

elisechant:


client/src/editor/scripts/redux/root/rootReducer.js
```diff
@@ -65,53 +75,82 @@ const rootReducer = reduceReducers(
...
-   recentDatapointIdx: null
-  },
-  {
-   // ...
-  },
-  // ...
- ]
- };
+
+  datagroupset - type:simple
```
elisechant: @micapam documenting datagroupset and datagroupsetSlice here fyi

PR #535, opened 2017-01-12
Coverage for both JS & Ruby via CodeCov (PD-999)

micapam:


circle.yml
```diff
@@ -31,7 +31,8 @@ dependencies:
 
 test:
   override:
-    - yarn test
+    - yarn run test:react:coverage
```
elisechant: perfect 👍

PR #542, opened 2017-01-16
test - api

elisechant:


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.3'
+ruby '2.3.1'
```
elisechant: dont commit this

client/src/test/utils/testUtils.js
```diff
@@ -1,9 +1,20 @@
 /* global jest */
 require('./../../../config/polyfills');
 
+import {isObject} from 'lodash';
```
nathf: To avoid importing the whole lodash lib this should be:
`import isObject from 'lodash/isObject'`

client/src/editor/scripts/components/reduxFormFields/inputHidden.js
```diff
@@ -6,9 +6,9 @@ import React from 'react';
  * @returns {Object<jsx>}
  * @constructor
  */
-const InputHidden = ({input}) => {
-  const {name} = input;
-  return <input {...input} type="hidden" id={name} />
+const InputHidden = (props) => {
+  const {input} = props;
```
nathf: This could have been done in the arg as well?

elisechant: yeah im not sure how to make doc blocks play nice with destructuring. Any ideas?

client/src/test/utils/testUtils.js
elisechant: yeah i need to do this basically EVERYWHERE also. fek.

client/src/editor/scripts/components/reduxFormFields/inputHidden.js
nathf: Guess it depends on your editor. It's only a minor thing (probably transpiled the same way anyway)

PR #548, opened 2017-01-17
PD-1005 - format currency and percentiles to 2dp in previews

elisechant:


client/src/editor/scripts/components/datagroupPreview.js
```diff
@@ -18,9 +19,18 @@ const Preview = ({recentDatagroupset}) => {
         let units = getHumanisedUnits(group.dataset.units);
         let unitsStr = units ? ` ${units}` : '';
         if (group.datapoint) {
```
elisechant: return null if value doesn't exist

client/src/editor/scripts/utils/formatPercentile.js
```diff
@@ -0,0 +1,11 @@
+
+/**
+ * Takes a value and returns a Number as String to 2 decimal places.
+ * Does not round.
+ * @param value {Number|String}
+ * @returns {string}
+ */
+export const formatPercentile2dp = value => {
```
nathf: Is the typing of this function intentionally flexible? Is it possible to be stricter allowing only Numbers?

elisechant: well it doesn't matter because Number won't be fussy about it?

nathf: If the string isn't a valid number for whatever reason

client/src/editor/scripts/components/datagroupPreview.js
```diff
@@ -12,15 +13,25 @@ const Preview = ({recentDatagroupset}) => {
...
 
       {recentDatagroupset.groups.map((group, idx) => {
-
         let value;
-
         let units = getHumanisedUnits(group.dataset.units);
         let unitsStr = units ? ` ${units}` : '';
         if (group.datapoint) {
-          value = group.datapoint.value === null ? 'No data' : group.datapoint.value + unitsStr;
+          if (group.datapoint.value === null) {
```
elisechant: should pull this out in to a testable unit of code

client/src/editor/scripts/components/datagroupPreview.js
```diff
@@ -1,33 +1,53 @@
...
+import {formatPercentile2dp} from './../utils/formatPercentile';
+
+
+/**
+ * Format and decorate value provided for the View
+ * @param value {Number|String<Number>}
+ * @param units {String}
+ * @returns {String}
+ */
+const formatValue = (value, units) => {
```
elisechant: write a test for this function

client/src/editor/scripts/utils/formatPercentile.js
elisechant: Ok. And then I can type check for that inside the function.

PR #549, opened 2017-01-17
testing, defensive coding and file renaming for urlHelpers

elisechant:


client/src/editor/scripts/utils/formatUrl.js
```diff
@@ -0,0 +1,73 @@
+
+export const getDashboardUrl = dashboardId => {
+  if (typeof dashboardId === 'undefined') {
+    if (process.env.NODE_ENV !== 'test') {
```
nathf: Whats the reason behind targeting `test`?

elisechant: because i'm testing the case typeof dashboardId === 'undefined' so it prints the console.errors in the test. Is there a better way to do this?

client/src/editor/scripts/utils/formatUrl.js
```diff
@@ -0,0 +1,73 @@
...
+      console.error('Parameters were not correctly provided', dashboardId);
+    }
+    return null;
+  }
+  return `/dashboards/${dashboardId}`;
+};
+
+export const getDashboardWidgetsUrl = dashboardId => {
+  if (typeof dashboardId === 'undefined') {
```
freshtonic: What's wrong with throwing an error instead? No point in trying to behave when something is broken. Scream loudly 😱💥

elisechant: hmm.. fair enough.

client/src/editor/scripts/utils/formatUrl.js
elisechant: maybe i should have a console utility that wraps that up behind the scenes rather than repeating it?

nathf: What about mocking out console.error with a spy (or similar) and ensuring that its called?

nathf: Or per @freshtonic comment below, ensure an error is being thrown.

client/src/editor/scripts/utils/console.js
```diff
@@ -0,0 +1,21 @@
+
+let win = typeof global === 'undefined' ? window : global;
+
+/**
+ * Explicit alias with additional powers for implied console.
+ * Use it just like window.console
+ */
+export default {
```
elisechant: @nathf thoughts?

nathf: Still prefer the spy method, it will capture the correct behaviour.

elisechant: 🤔  ok.

client/src/editor/scripts/utils/formatUrl.test.js
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  afterEach(function() {
+    spy.restore();
+  });
+
+  describe('getDashboardUrl', () => {
+    it('should return null if no param is provided', () => {
+      expect(getDashboardUrl()).toBe(null);
+      expect(spy).toHaveBeenCalled();
```
elisechant: great idea! thanks @nathf

